Issue #3, opened 2022-10-07
Authenticate With Garden Client


BenGalewsky:
_As a Garden Publisher I want to authenticate with Globus so I can have permission to publish my model_

# Assumptions:
1. There will be a Garden globus app
2. The constructor will obtain tokens via GlobusAuth3. 
3. Tokens will be persisted in the user's `.garden` directory 
4. Story is burdened with setting up readthedocs 

# Acceptance Criteria
1. Given I have garden installed in my python environment, and it is the first time I've used the GardenClient when I execute python code to instantiate a GardenClient, Then I will be presented with a link to authenticate and an opportunity to paste a token and it will be accepted.
2. Given I have garden installed in my python environment and I've successfully authenticated the GlobusClient, when I execute python code to instantiate a GardenClient then it will not ask for me to follow the link and paste a token

# Tasks
- [x] Create Garden Application in [https://auth.globus.org/v2/web/developers](https://auth.globus.org/v2/web/developers)
- [x] Update project to use poetry and include mypy and safety in CI (PR #12 )
- [x] Code up auth workflow in client constructor (PR #13)
- [x] Code up unit testing strategy for auth as well as testing out authenticated methods (PR #13)
- [x] Create readthedocs flow with Sphinx api documentation

BenGalewsky:
@blaiszik - I assume that you should own the Garden application on globus auth? Also, do we have a secure way to share the secrets? My go-to would be lastpass

BenGalewsky:
@ericblau do you have a terrific example of doing Globus Auth from a python client? I want our implementation to be a textbook example.

I looked at the [funcX client](https://github.com/funcx-faas/funcX/blob/fe54444868e7393d60be03db4c1b426df32a2d51/funcx_sdk/funcx/sdk/client.py#L151) and it seems incredibly complicated - I don't think we need anything that sophisticated!

BenGalewsky:
> Do we need to share the credentials, or could we just have them in GH Secrets?

You typically need to know them for development. On another project we kept passing them around on slack which is not particularly great.

blaiszik:
They are now available through CLIENT_ID and CLIENT_SECRET in GitHub secrets

ericblau:
[...]

PR #13, opened 2022-10-28
Add Authentication to Garden Client

BenGalewsky:
# Problem
Users of the GardenClient need to obtain valid GlobusAuth tokens in order to access the service.

Solution to #3 

# Approach
1. Added Globus-SDK to project.
2. Used NativeAppAuthorizer to launch an OAuth3 flow to obtain a token.
3. Use the JSON TokenStore to save tokens.
4. Created unit tests to exercise the auth workflow

This story includes setting up Sphinx documentation and publishing to readthedocs.

Notes:
There is a [ridiculous pin in flake8 on an old version of importlib-metadata](https://github.com/PyCQA/flake8/pull/1438) which is incompatible with Sphinx. I had hoped to put sphinx tooling in `pyroject.yaml`, but it meant pinning to a very old version of flake8. In the end I kept flake8 and removed the sphinx libraries.

### DIFF
--- a/.github/workflows/documentation-links.yaml
+++ b/.github/workflows/documentation-links.yaml
@@ -1,0 +1,18 @@
+# .github/workflows/documentation-links.yaml
+
+name: Read the Docs Pull Request Preview
+on:
+  pull_request_target:
+    types:
+      - opened
+
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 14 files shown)


ericblau (approved):
This all looks good to me.  We might consider adding opening a local server to eliminate the need to cut and paste codes (if the client is being run locally), ala https://github.com/globus/native-app-examples/blob/main/example_local_server.py, but with local token storage and refresh tokens, cutting and pasting codes should be rare anyway.

PR #20, opened 2022-11-13
Update Version

Aadit-Ambadkar:
Minor update since we are on version 0.0.2 in pypi.

----
:books: Documentation preview :books:: https://garden-ai--20.org.readthedocs.build/en/20/

### DIFF
--- a/garden_ai/version.py
+++ b/garden_ai/version.py
@@ -1,3 +1,3 @@
 # single source of truth for package version,
 # see https://packaging.python.org/en/latest/single_source_version/
-__version__ = "0.0.1"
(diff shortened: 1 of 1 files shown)


Aadit-Ambadkar:
> I'd suggest deleting `version.py` - to make library publishing easy via GitHub release workflow, the GitHubAction picks up the released version number from the workflow and [sets it dynamically at release time](https://github.com/Garden-AI/garden/blob/6ffd4c6f57d4b52039526d887d57e2dd50df33ef/.github/workflows/pypi.yaml#L22).

sounds good.

PR #21, opened 2022-11-17
4 garden metadata creation

### LINKED ISSUE #4: Garden metadata creation
_As a Garden Publisher I want to create garden metadata so consumers can easily find my model_

# Assumptions
1. Metadata will include:
 a. Model title
 b. Model description
 c. DataCite information about the submission
2. Data will be defined using [pydantic](https://pydantic-docs.helpmanual.io)
3. The GardenClient interface will expose a [fluent API](https://en.wikipedia.org/wiki/Fluent_interface) starting with a `create_model` method.
4. This story ends with the library writing a JSON file containing this metadata to the user's current working directory 

#Acceptance Criteria
Given I have instantiated a `GardenClient` when I call the `create_garden` with chained metadata setting methods specifying all of the required fields then I will find a JSON metadata file in my current working directory.

Given I have instantiated a `GardenClient` when I call the `create_garden` with chained metadata setting methods, committing a required field then I will receive an exception.

OwenPriceSkelly:
(Tentatively) this closes #4 

----
:books: Documentation preview :books:: https://garden-ai--21.org.readthedocs.build/en/21/

### DIFF
--- a/garden_ai/garden.py
+++ b/garden_ai/garden.py
@@ -2,8 +2,13 @@ import logging
 import os
 from pathlib import Path
 
-from globus_sdk import GroupsClient, RefreshTokenAuthorizer, NativeAppAuthClient, \
-    AuthClient, AuthAPIError
+from globus_sdk import (
+    GroupsClient,
+    RefreshTokenAuthorizer,
+    NativeAppAuthClient,
+    AuthClient,
+    AuthAPIError,
+)
 from typing import List
 from globus_sdk.tokenstorage import SimpleJSONFileAdapter
 
@@ -33,12 +38,16 @@ class GardenClient:
     def __init__(self, auth_client: AuthClient = None):
         key_store_path = Path(os.path.expanduser("~/.garden"))
         key_store_path.mkdir(exist_ok=True)
-        self.auth_key_store = SimpleJSONFileAdapter(os.path.join(key_store_path, "tokens.json"))
-        self.client_id = os.environ.get("GARDEN_CLIENT_ID", "cf9f8938-fb72-439c-a70b-85addf1b8539")
+        self.auth_key_store = SimpleJSONFileAdapter(
+            os.path.join(key_store_path, "tokens.json")
+        )
+        self.client_id = os.environ.get(
+            "GARDEN_CLIENT_ID", "cf9f8938-fb72-439c-a70b-85addf1b8539"
+        )
 
-        self.auth_client = NativeAppAuthClient(self.client_id) \
-            if not auth_client\
-            else auth_client
+        self.auth_client = (
+            NativeAppAuthClient(self.client_id) if not auth_client else auth_client
+        )
 
         self.authorizer = self._authenticate()
 
@@ -83,7 +92,7 @@ class GardenClient:
         )
         return authorizer
 
-    def create_garden(self, creators: List[str] = [], title: str = "", **kwargs):
+    def create_garden(self, authors: List[str] = [], title: str = "", **kwargs):
         """Construct a new Garden object, optionally populating any number of metadata fields from `kwargs`.
 
         Up to user preference, metadata (e.g. `title="My Garden"` or
@@ -98,10 +107,10 @@ class GardenClient:
 
         Parameters
         ----------
-        creators : List[str]
+        authors : List[str]
             The personal names of main researchers/authors involved in
             cultivating the Garden. Names should be formatted "Family, Given",
-            e.g. `creators=["Mendel, Gregor"]`. Affiliations/institutional
+            e.g. `authors=["Mendel, Gregor"]`. Affiliations/institutional
             relationships should be added via the Garden object helper method
             `add_affiliation`, e.g.  `pea_garden.add_affiliation({"Mendel,
             Gregor": "St Thomas' Abbey"})`. (NOTE: add_affiliation not yet implemented)
@@ -119,7 +128,7 @@ class GardenClient:
         --------
             gc = GardenClient()
             pea_garden = gc.create_garden(
-                creators=["Mendel, Gregor"],
+                authors=["Mendel, Gregor"],
                 title="Experiments on Plant Hybridization",
                 subjects=["Peas"]
             )
@@ -127,8 +136,8 @@ class GardenClient:
             pea_garden.subjects += ["Genetics"] # (didn't have the word for it earlier)
         """
         data = dict(kwargs)
-        if creators:
-            data["creators"] = creators
+        if authors:
+            data["authors"] = authors
         if title:
             data["title"] = title
         return Garden(**data)
--- a/garden_ai/garden.py
+++ b/garden_ai/garden.py
@@ -4,8 +4,11 @@ from pathlib import Path
 
 from globus_sdk import GroupsClient, RefreshTokenAuthorizer, NativeAppAuthClient, \
     AuthClient, AuthAPIError
+from typing import List
 from globus_sdk.tokenstorage import SimpleJSONFileAdapter
 
+from garden_ai.model import Garden
+from pydantic import ValidationError
 
 logger = logging.getLogger()
 
@@ -79,3 +82,87 @@ class GardenClient:
             on_refresh=self.auth_key_store.on_refresh,
         )
         return authorizer
+
+    def create_garden(self, creators: List[str] = [], title: str = "", **kwargs):
+        """Construct a new Garden object, optionally populating any number of metadata fields from `kwargs`.
+
+        Up to user preference, metadata (e.g. `title="My Garden"` or
+        `version="1.0.0"`) can be provided here as kwargs.
+
+        This might be useful if, for example, one wanted to build a Garden starting
+        from an already-existing Garden or pre-populated dict of template
+        metadata. Otherwise, the user is free to incrementally populate or
+        replace even the Garden object's required fields (e.g. `pea_garden.title
+        = "Experiments on Plant Hybridization"`) at any time -- field validation
+        is still performed.
+
+        Parameters
+        ----------
+        creators : List[str]
+            The personal names of main researchers/authors involved in
+            cultivating the Garden. Names should be formatted "Family, Given",
+            e.g. `creators=["Mendel, Gregor"]`. Affiliations/institutional
+            relationships should be added via the Garden object helper method
+            `add_affiliation`, e.g.  `pea_garden.add_affiliation({"Mendel,
+            Gregor": "St Thomas' Abbey"})`. (NOTE: add_affiliation not yet implemented)
+
+        title : str
+            An official name or title for the Garden. This attribute must be set
+            in order to register a DOI.
+
+        **kwargs :
+            Metadata for the new Garden object. Keyword arguments matching
+            required or recommended fields will be (where necessary) coerced to the
+            appropriate type and validated per the Garden metadata spec.
+
+        Examples
+        --------
+            gc = GardenClient()
+            pea_garden = gc.create_garden(
+                creators=["Mendel, Gregor"],
+                title="Experiments on Plant Hybridization",
+                subjects=["Peas"]
+            )
+            pea_garden.year = 1863
+            pea_garden.subjects += ["Genetics"] # (didn't have the word for it earlier)
+        """
+        data = dict(kwargs)
+        if creators:
+            data["creators"] = creators
+        if title:
+            data["title"] = title
+        return Garden(**data)
+
+    def register_metadata(self, garden: Garden, out_dir=None):
+        """Make a Garden object's metadata discoverable.
+
+        This will perform validation on the metadata fields and (if successful)
+        write the metadata to a `"metadata.json"` file in the current working directory.
+
+        Parameters
+        ----------
+        garden : Garden
+            A Garden object with user's ready-to-publish metadata. Users might
+            want to invoke `to_do()` method on their Garden, or see `Garden`
+            docstring for explanation of any unset required/recommended fields.
+
+        out_dir : Union[PathLike, str]
+            Directory in which a copy of the Garden's metadata is written on
+            successful registration. Defaults to current working directory.
+
+        Raises
+        ------
+        ValidationError
+
+        """
+        out_dir = Path(out_dir) if out_dir else Path.cwd()
+        out_dir /= "metadata.json"
+        try:
+            garden.request_doi()
+            garden.validate()
+        except ValidationError as e:
+            logger.error(e)
+            raise
+        else:
+            with open(out_dir, "w+") as f:
+                f.write(garden.json())
--- a/garden_ai/model.py
+++ b/garden_ai/model.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+import logging
+from datetime import datetime
+from typing import List
+from uuid import uuid4, UUID
+from pydantic import BaseModel, Field, validator, ValidationError
+
+logger = logging.getLogger()
+
+
+class Garden(BaseModel):
+    """Object representation of a Garden™.
+
+    Required Attributes
+    --------------------
+    authors: List[str]
+        The main researchers involved in producing the Garden. At least one
+        author is required in order to register a DOI. Personal name format
+        should be: "Family, Given".
+
+    title: str
+        An official name or title for the Garden. This attribute must be set
+        in order to register a DOI.
+
+
+    Recommended Attributes
+    --------------------
+    description: str
+        A brief summary of the Garden and/or its purpose, to aid discovery by
+        other Gardeners.
+
+
+    Optional Attributes
+    --------------------
+    doi: str
+        A garden's doi can be manually set by the user, or generated automatically via the
+        DataCite rest api using the required fields.
+
+    language: str
+        Recommended values are IETF BCP 47, ISO 639-1 language codes, such as:
+        "en", "de", "fr". Defaults to "en".
+
+    subjects: List[str]
+        Subjects, keywords, classification codes, or key phrases pertaining to the Garden.
+
+    year: str
+        Defaults to current year. This attribute should be formatted 'YYYY'.
+
+    Examples
+    --------
+    Creating a new Garden with a ``GardenClient`` instance::
+
+        gc = GardenClient()
+        pea_garden = gc.create_garden(
+            authors=["Mendel, Gregor"], title="Experiments on Plant Hybridization"
+        )
+        pea_garden.year = 1863
+        pea_garden.description = '''This Garden houses sophisticated ML pipelines
+                                  for Big Pea Data extraction and classification.
+                                  It consists of a 2-hectare plot behind the monastery,
+                                  and a 30,000-plant dataset.'''
+
+        gc.register(pea_garden)
+
+    Notes
+    --------
+    Mendel's work was ignored by the scientific community during his lifetime,
+    presumably due to the lack of a working DOI.
+    To remedy this, if the doi field is unset when registering the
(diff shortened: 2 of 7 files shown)


garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
             on_refresh=self.auth_key_store.on_refresh,
         )
         return authorizer
+
+    def create_garden(self, **kwargs):
```
BenGalewsky, line 85: My vision was to make the required fields named parameters in this constructor so it's clear in your IDE exactly what is the minimum for creating a Garden.

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+
+            gc = GardenClient()
+            pea_garden = gc.create_garden(
+                creators=["Mendel, Gregor"], title="Experiments on Plant Hybridization"
+            )
+            pea_garden.year = 1863
+        """
+        return Garden(**kwargs)
+
+    def register_garden(self, G: Garden):
```
BenGalewsky, line 113: Capital letter for this parameter seems non-pythonic, right?

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+
+from datetime import datetime
+from enum import Enum, auto
+from typing import List
+from uuid import uuid4, UUID
+import pydantic
+from pydantic import BaseModel, Field, validator
+
+
+class TempField(Enum):
```
BenGalewsky, line 13: This seems complicated... what is going on here that's not just built into pydantic?

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+        def request_doi(garden):
+            # TODO
+            # to be replaced with code to build the actual doi
+            return "10.55555/fake-doi"
+        try:
+            # attempts to populate a dummy garden with the data from the real
+            # one, so that it invokes our custom validators on every field (in
+            # particular, those which still have TempField defaults)
+            G.doi = request_doi(G)
+            _ = Garden(**G.dict())
```
BenGalewsky, line 137: Might was well make `validate` a first class method on `Garden` - others may want to call this directly

pyproject.toml
```diff
@@ -28,6 +29,16 @@ safety = "^2.3.1"
 sphinx = "4.3.2"
 sphinx-rtd-theme = "^1.0.0"
 
+
+[tool.poetry.group.dev]
+optional=true
+
+[tool.poetry.group.dev.dependencies]
+python-lsp-server = {extras = ["rope"], version = "^1.6.0"}
```
BenGalewsky, line 37: Hmm, does anyone else use these dependencies? I wonder how the team feels about adding such specific development tooling to our common pyproject?

BenGalewsky:
Overall this is a really great start! You've made a very abstract set of requirements concrete. 
I've added some specific comments. More generally, you will need to add unit tests to keep our [code coverage](https://app.codecov.io/gh/Garden-AI/garden/tree/4-garden-metadata-creation) up.

garden_ai/garden.py
BenGalewsky, line 85: Also wonder if this should just be a Garden constructor - not sure if having this as a method on GardenClient adds anything

garden_ai/garden.py
ascourtas: agreed, also full-named parameters are preferred to abbreviations.

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
             on_refresh=self.auth_key_store.on_refresh,
         )
         return authorizer
+
+    def create_garden(self, **kwargs):
+        """Initialize and return a new Garden object
```
ascourtas, line 86: some more clarity in the docstring on what a Garden object is would be helpful

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+
+        If invoked with keyword arguments corresponding to field names (e.g.
```
ascourtas, line 88: this sentence is a little confusing, please clarify

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+        `title` or `creator`) are passed directly to the `Garden` model, making
+        it possible for the user to provide necessary info upfront to be parsed
+        and validated at initialization, or to incrementally provide field
+        values via simple assignment to attributes of the returned object
+        directly.
+
+        Parameters
+        ----------
+        **kwargs :
+            Metadata for the new Garden object. Keyword arguments matching
```
ascourtas, line 98: how will a user know what these kwargs should be? please add more context

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+            required or recommended fields will be coerced to the appropriate
+            type, if necessary.
+
+        Examples
```
ascourtas, line 102: love the example! <3

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+            return "10.55555/fake-doi"
+        try:
+            # attempts to populate a dummy garden with the data from the real
+            # one, so that it invokes our custom validators on every field (in
+            # particular, those which still have TempField defaults)
+            G.doi = request_doi(G)
+            _ = Garden(**G.dict())
+        except ValidationError as e:
+            logger.error(str(e))
+            print(e)
```
ascourtas, line 140: please remove print statements

garden_ai/garden.py
ascourtas, line 137: please add an inline comment about what is being accomplished here. the existing `Garden` metadata is being unpacked to create a new `Garden` object? feels funny

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+            gc = GardenClient()
+            pea_garden = gc.create_garden(
+                creators=["Mendel, Gregor"], title="Experiments on Plant Hybridization"
+            )
+            pea_garden.year = 1863
+        """
+        return Garden(**kwargs)
+
+    def register_garden(self, G: Garden):
+        """Finalize the setup of a Garden object with user-populated fields
```
ascourtas, line 114: please add more context. What does "finalize" mean? The method name has the word "register" -- what is being registered here? from the code, it looks like the key component is the minting of the DOI. Please make that clear in the docstring

garden_ai/garden.py
OwenPriceSkelly: I like that idea -- we could also just expose the `Garden` constructor to the user, because my impression is that pydantic makes a point of playing nicely with IDEs/etc

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+        try:
+            # attempts to populate a dummy garden with the data from the real
+            # one, so that it invokes our custom validators on every field (in
+            # particular, those which still have TempField defaults)
+            G.doi = request_doi(G)
+            _ = Garden(**G.dict())
+        except ValidationError as e:
+            logger.error(str(e))
+            print(e)
+            raise TypeError
```
ascourtas, line 141: if you are raising a TypeError, I suggest wrapping it with some context so the user sees the logged error more easily. Something like the following, but with more context (hard for me to tell from this method):

```
raise TypeError(f"Error while registering Garden: {e.message}")
```

garden_ai/garden.py
OwenPriceSkelly: Oh, but that would be providing completion suggestions for all the fields, not *only* the required fields, which I think is what you were getting at

garden_ai/garden.py
OwenPriceSkelly: indeed

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+        else:
+            outfile = Path.cwd() / 'metadata.json'
```
ascourtas, line 143: could be good to log to `info` that a file is being written locally?

ascourtas, line 143: not immediately clear what is happening in this line, please add inline comment

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+    REQUIRED = auto()  # just those needed to get a doi
+    WARN = auto()
+    # setting this should trigger warning before publishing the model
+    # we can also inspect the object to see which fields have been set manually
+    # might be more flexible in the future
+
+
+class Garden(BaseModel):
+    """Object representation of a Garden™.
```
ascourtas, line 22: lol the little TM ✨

garden_ai/garden.py
BenGalewsky: It would be something like `create_garden(title: str, authors: list[str], **kwargs)` 

So the mandatory fields would be obvious, and the rest would be keywords

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+        other Gardeners.
+
+
+    Optional Attributes
+    ----------
+    doi: str
+        A garden's doi can be manually set by the user, or generated automatically via the
+        DataCite rest api using the required fields.
+
+    language: str
```
ascourtas, line 50: Nice, but add clarity on what this means -- I assume it's just a metadata field explaining what language the abstract, etc are written in?

Otherwise, it could be read as a field for setting the language (ie we do translation)

garden_ai/model.py
ascourtas: Please add a little context about what a Garden is

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+
+    subjects: List[str]
+        Subjects, keywords, classification codes, or key phrases pertaining to the Garden.
+
+    Examples
+    --------
+    Creating a new Garden with a ``GardenClient`` instance::
+
+        gc = GardenClient()
+        pea_garden = gc.create_garden(
```
ascourtas, line 62: <3

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+                                  for Big Pea Data extraction and classification.
+                                  It consists of a 2-hectare plot behind the monastery,
+                                  and a 30,000-plant dataset.'''
+
+        gc.register(pea_garden) # we can also have this invoke a pydantic "root validator" we define elsewhere
+
+    Notes
+    --------
+    Mendel's work was ignored by the scientific community during his lifetime,
+    presumably due to the lack of a working DOI.
```
ascourtas, line 77: 😂 amazing. High quality work.

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+        validate_assignment = True  # validators invoked (only?) on assignment
+
+    creators: List[str] = Field(TempField.REQUIRED, min_items=1, unique_items=True)
+    title: str = Field(TempField.REQUIRED)
+    resourceTypeGeneral: str = "Other"  # (or: model, software, service, interactive?)
+    publisher: str = "Garden"
+    year: str = Field(default_factory=lambda: str(datetime.now().year))
+
+    # dois currently only validated against this regex
+    # doi: str = Field(TempField.REQUIRED, alias="identifier", regex=r"^10.\d{4,9}/[-._;()/:A-Z0-9]+$")
```
ascourtas, line 100: is the commented out code in this Config meant to be here?

poetry.lock
```diff
@@ -31,6 +31,29 @@ python-versions = ">=3.6"
 [package.dependencies]
 pytz = ">=2015.7"
 
+[[package]]
```
ascourtas, line 34: are you supposed to commit the lock file? This was contentious for Pipenv in its time

ascourtas (commented):
Great job Owen! This is good quality code. I especially loved the examples. I added some comments on structure and readability/understandability. 

I didn't comment on the Pydantic class defs as much because I am less familiar -- I will leave that for @WardLT ✨

garden_ai/garden.py
ascourtas: I was also thinking this was a little funny, but Owen's example made the logical user flow really nice? I am torn.

WardLT (commented):
I like the mechanism for marking recommended fields, but think we should have it occur when a model is being exported (e.g., via `dict()` or `json()`) so that users don't get warnings that their object is incomplete before finishing. 

Maybe write "check recommended fields" as a function and overload the `dict`/`json` methods to run it first before export. 

We could also look up the field description in the `__fields__` attribute of the class so that we can print out a "what is this field" in the warning. 

How does that sound?

pyproject.toml
OwenPriceSkelly: I was also conflicted about this, the lsp tooling I use was deeply confused by poetry and this seemed to be the workaround -- do you know if there's a less obnoxious way for me to include those as dev dependencies? I could also just not commit those changes to pyproject.toml on my end, but that feels like asking for trouble

garden_ai/garden.py
OwenPriceSkelly: I agree with both points, I think @BenGalewsky's vision earlier about `create_garden(title: str, authors: list[str], **kwargs)` is a nice compromise

garden_ai/model.py
BenGalewsky: I wonder if DOI will be set in the client or added in the server anyway. If that's the case then we have our 100% required fields, which are set in the constructor.. everything else is 100% optional, so we don't need any fancy footwork.

Even if the DOI is fetched in the client, then it's not technically a required field. We would just go set it if it's empty and carry on. I think we building extra complexity that is not actually required

garden_ai/garden.py
OwenPriceSkelly: yeah I agree it does feel funny. that's what the few lines of comments above are failing  to totally explain -- the way the pydantic model is currently set up, we're allowed to build a blank object that has invalid default values (taken from the `TempField` enum in model.py) but only checks for validity when something tries to explicitly set them. 

So if someone has a Garden they _think_ they've finished but they never gave a title, this would look to pydantic like someone tried to construct a dummy Garden that explicitly sets `title=TempField.REQUIRED`, which would implicitly trigger whatever helpful pydantic errors/exceptions we want it to

tldr: one-liner that forces a call to every validator, in case there are still fields that haven't been validated yet

garden_ai/garden.py
OwenPriceSkelly: this is the most placeholder-y part of the code I think. my reasoning here was (copied from my convo with Logan in slack): if a user is invoking some `GardenClient` method to finalize/register their `Garden` object, that's a safe point to assume the user is "done" (or thinks they're done) providing metadata and seemed like a reasonable point to run a full suite of validation functions

garden_ai/garden.py
ascourtas: That is an excellent question that I think is also tied to the above comment about whether or not `create_garden()` should just be a regular constructor for a class. 

In the past, we have used names like `FoundryMetadata` or `FoundryBaseModel` to describe Pydantic classes describing the metadata, where then `Foundry` might be the class that describes the actual Foundry object. 

I am not as familiar with Pydantic, so my suggestion would be to read into what the best practices are for how to conceptually organize this to be in-line with OOP principles. 

A major question here is we know that the Pydantic class describes the state (ie attributes) of a Garden, but what about the function? Does that go in another `Garden` class, or just in the `GardenClient`? 

If there are no clear answers, I'd lean towards the Pydantic class being called `GardenMetadata` or something similar, and then having the actual object (which contains `GardenMetadata` as an attribute) be `Garden` or `GardenClient`, depending on where we want function to live.

garden_ai/garden.py
OwenPriceSkelly: I think that design-wise this method is probably trying to do too much, and to be fully transparent I was kinda hoping "register" would be evocative but vague enough to not think too hard about what it meant

garden_ai/model.py
OwenPriceSkelly: the regex convenience option that pydantic provided there was causing later validation for the doi to skip for some reason, I'm not sure if it was a pydantic bug or a me bug and I thought I might revisit it later, but also "I'll revisit this later" could be famous last words so I can just take it out

garden_ai/garden.py
OwenPriceSkelly: oh I left this one in on purpose but I meant to ask if it should be a log instead -- this print statement was there to display the pydantic Helpful Error Message to the user the way they do in the pydantic docs examples

garden_ai/model.py
```diff
@@ -0,0 +1,146 @@
...
+    garden_id: UUID = Field(default_factory=uuid4, allow_mutation=False)
+
+    # always=False should prevent this from being called until an attempt to set
+    # a field, but this should be called for every field.
+    @validator("*", pre=True, always=False)
+    def check_required(cls, val, field: pydantic.fields.ModelField):
+        assert val is not TempField.REQUIRED, f"{field.name} must be set"
+        if val is TempField.WARN:
+            # TODO: log warning
+            print(f"Warning: {field.name} is recommended but unset")
```
Aadit-Ambadkar, line 118: Wouldn't it make more sense to do logging instead of print statements?

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+        TypeError
+
+        Examples
+        --------
+        TODO
+        """
+        def request_doi(garden):
+            # TODO
+            # to be replaced with code to build the actual doi
+            return "10.55555/fake-doi"
```
Aadit-Ambadkar, line 131: Consider making this a publicly available function. The user may want access to this.

garden_ai/garden.py
Aadit-Ambadkar: Use logging instead of print statements.

garden_ai/garden.py
Aadit-Ambadkar: 👍🏼 agreed

garden_ai/garden.py
```diff
@@ -79,3 +81,65 @@ def _authenticate(self):
...
+            G.doi = request_doi(G)
+            _ = Garden(**G.dict())
+        except ValidationError as e:
+            logger.error(str(e))
+            print(e)
+            raise TypeError
+        else:
+            outfile = Path.cwd() / 'metadata.json'
+            with open(outfile, 'w+') as f:
+                f.write(G.json())
```
Aadit-Ambadkar, line 145: what's the purpose of this? (this is a geniuine question, not a review comment)

garden_ai/model.py
blaiszik: yes, good idea Aadit!

garden_ai/garden.py
OwenPriceSkelly: Context so I can resolve this in good conscience: the current `Garden` object is on track to be greater than the sum of its metadata, with more functionality than would be provided by a more vanilla pydantic model. This feels ok to me, but I do think you're right here to question whether this is good OOP, & my feelings won't be hurt if we later decide that the interface would be better if garden metadata and ~gardening functionality were owned by different classes

garden_ai/garden.py
OwenPriceSkelly: Made it clearer in docstring that this is mainly for tying up loose metadata-ends

garden_ai/garden.py
OwenPriceSkelly: agreed, ended up attaching the method to the Garden object, but might make sense in the future to move it to the garden client object

garden_ai/garden.py
OwenPriceSkelly: just to hit the acceptance criteria spelled out in #4

Issue #15, opened 2022-10-30
Publish Model Metada


BenGalewsky:
_As a model publisher I want to publish my model metadata so consumers can easily find my model_

# Assumptions:
1. Metadata will be published in GlobusSearch
2. The publication will occur from the client and not require an intermediate server.3. 
3. Writes to GlobusSearch will be controlled by membership in a GlobusGroup

# Acceptance Criteria
Given I have described my models metadata using the GardenClient, when I call the `publish` method, then I can find my metadata on GlobusSearch using `globus-search-cli`

blaiszik:
Following this model anyone in the Globus Group will be able to write or overwrite anything in the index. This is ok as an incremental step, but just a reminder of the way permissions work for Globus Search indexes.

BenGalewsky:
> Following this model anyone in the Globus Group will be able to write or overwrite anything in the index.

Can this only be fixed by having our own server in the middle?

ericblau:
I was going to assign myself to this, as I know I have all necessary repository permissions to work on it, but I don't have permissions to edit the assignee field

Aadit-Ambadkar:
> > Following this model anyone in the Globus Group will be able to write or overwrite anything in the index.
> 
> Can this only be fixed by having our own server in the middle?

You could hash a password for each model (when the model is created, the publisher passes in a password which is hashed and stored) and when a user wants to edit the metadata they pass in the password and edit it.

ericblau:
> > Can this only be fixed by having our own server in the middle?
> 
> You could hash a password for each model (when the model is created, the publisher passes in a password which is hashed and stored) and when a user wants to edit the metadata they pass in the password and edit it.

If all of the checks are client-side, a malicious or motivated user could bypass them.
We could absolutely devise a method by which we allow the user to modify (only) their own documents--the simplest way would be to add a "published by" field to the doc and stick their globus uuid (and/or username, but uuid is probably better) in it.  Then, the server can check the existing doc against your uuid before letting you republish.  

  It's a bit annoying that the Globus Search model lets you set accessibillity (visibility) on a per-document basis, but write permission at a per index basis.  But that's what we've got.  (and dynamoDB wouldn't be any better out of the box--we'd still have to roll our own access control, server side).  

The solution is to have a server do all of the actual publishing, and not give index write permissions to any normal users.  The server then does the check as to whether the user should be able to republish the doc (as above, or with a hashed password as you suggest). That's out of scope for this story, though.

[...]

PR #22, opened 2022-11-29
Globus Search publishing of Garden metadata

ericblau:
This PR implements a method to publish Garden metadata to a Globus Search search index, as discussed in issue #15  

----
:books: Documentation preview :books:: https://garden-ai--22.org.readthedocs.build/en/22/


garden_ai/garden.py
```diff
@@ -175,3 +212,26 @@ def register_metadata(self, garden: Garden, out_dir=None):
         else:
             with open(out_dir, "w+") as f:
                 f.write(garden.json())
+
+    def publish_garden(self, garden_id=None, garden_doc=None, visibility='Public'):
```
OwenPriceSkelly, line 216: is `garden_doc` supposed to be the metadata dict for the garden object? if so I think it might make more sense to have this just take a `garden` and use `garden.uuid` and `dict(garden.json())` to get the garden id and metadata dict respectively

ericblau: Yeah, when I first wrote this, the garden object didn't exist yet.  I totally agree that it should just take a garden--I'll make the change.

PR #34, opened 2023-01-12
implement `request_doi` method for gardens

OwenPriceSkelly:
_as a Gardener, I want my `Garden` object to be able to register a DOI automatically, if I've provided enough other metadata to do so (and don't already have a DOI explicitly set)_

# Assumptions:
1. the current `Garden` implementation defines a `__doi_required__` class attribute with the field names required to register for a DOI using the [DataCite api](https://support.datacite.org/docs/api-create-dois).
2. the current `Garden` has the first part of this already implemented in the `request_doi` method, which checks the `__doi_required__` fields but just returns "10.26311/fake-doi" on success, rather than actually hitting the DataCite API.
3. this should hit the DataCite test environment here: https://doi.test.datacite.org/ 
4. The Globus test DataCite id/password can be environment variables, not hardcoded

Note also that this is going to require some changes to existing unit tests

# Acceptance Criteria
Given a `Garden` or `Pipeline` with explicit values for `authors`, `title` and `year` fields  when `garden_client._mint_doi(obj: Garden | Pipeline)` is called, then the DataCite api should be used to mint a doi (via the garden backend) to set the `doi` field, else any DataCite-side errors should be propagated back to the user. 

When `garden_client.register_metadata(my_garden)`, then the garden _and any of its pipelines_ missing DOIs will request one before metadata is made "discoverable" (writing `garden.json`)

successfully authenticates with the backend, even if the backend does not yet have the expected environment variables set up with the datacite credentials 

----
:books: Documentation preview :books:: https://garden-ai--34.org.readthedocs.build/en/34/


BenGalewsky:
@OwenPriceSkelly is there an issue for this? We should start with an issue before the PR

OwenPriceSkelly:
@BenGalewsky yeah there was (same content as the first part of the convo here), not sure where it went -- I tried out a "make PR from issue" feature from my editor I hadn't tried before and it looks like that deleted the issue and made a PR instead?

OwenPriceSkelly:
is #44, so: closes #44

garden_ai/utils.py
```diff
@@ -196,3 +201,105 @@ def f_of_g(*args, **kwargs):
...
+
+    Raises
+    ------
+    HTTPError
+        if one is encountered by `requests.Response.raise_for_status`
+
+    NotImplementedError
+        if given `test=False`
+    """
+    if not test:
```
WillEngler, line 247: To the extent possible, it's nice to avoid having application code be aware that it's under test. I would make the doi prefix and datacite urls environment variables. There could be local default settings where they are the test values. A prod deployment would have them set to the real values.

garden_ai/utils.py
```diff
@@ -196,3 +201,105 @@ def f_of_g(*args, **kwargs):
...
+                },
+                "url": "http://thegardens.ai",
+                "schemaVersion": "http://datacite.org/schema/kernel-4",
+            },
+        }
+    }
+
+    try:
+        DATACITE_REPOSITORY_ID = os.environ["DATACITE_REPOSITORY_ID"]
+        DATACITE_PASSWORD = os.environ["DATACITE_PASSWORD"]
```
WillEngler, line 285: This is something that's awkward for client vs backend, so consider this a note for the second pass of this that lives in the backend. (And is quite possibly something you're already thinking about)

Ideally when there's something wrong with the app config, you want to fail as quickly as possible. When this gets ported to the backend, this should trigger some alarms in our logs at app startup so that we can know there's an issue ASAP (even if a user doesn't hit this part of the code)

WillEngler (approved):
Looking good! I'm not super locked in on functionality requirements yet so I'm leaning on other reviewers to say that the behavior is right. But I am leaving some operations-y suggestions. It would make sense if you don't want to address these until the backend draft.

OwenPriceSkelly:
current status of this is very briefly blocked -- we need to tweak the `requested_scopes` in the client to have things play nice with the new backend, and the backend will need to set the relevant `DATACITE_*` environment variables before the integration tests are going to pass again

garden_ai/garden.py
```diff
@@ -1,27 +1,35 @@
...
+from garden_ai.models import Garden, Pipeline
+
 logger = logging.getLogger()
 
 
 class AuthException(Exception):
     pass
 
 
+GardenScopes = ScopeBuilder(
```
OwenPriceSkelly, line 28: @ any reviewer(s) who would know, please double check that this (the `GardenScopes` class) is roughly the way to go about this? I wasn't familiar with the idiom before stumbling on analogous code in the globus sdk

garden_ai/garden.py
```diff
@@ -59,13 +69,20 @@ def __init__(self, auth_client: AuthClient = None, search_client: SearchClient =
...
         self.search_client = (
-           SearchClient(authorizer=self.searchauthorizer) if not search_client else search_client
+            SearchClient(authorizer=self.searchauthorizer)
+            if not search_client
+            else search_client
         )
+        self.garden_authorizer = self._create_garden_authorizer()
 
     def _do_login_flow(self):
+        print(GardenClient.scopes.action_all)
```
WillEngler, line 79: Probably meant to clean this up?

WillEngler (approved):

garden_ai/garden.py
```diff
@@ -1,26 +1,27 @@
...
 from globus_sdk.tokenstorage import SimpleJSONFileAdapter
-
-from garden_ai.models import Garden
 from pydantic import ValidationError
 
+from garden_ai.models import Garden, Pipeline
 
 # garden-dev index
-GARDEN_INDEX_UUID = '58e4df29-4492-4e7d-9317-b27eba62a911'
+GARDEN_INDEX_UUID = "58e4df29-4492-4e7d-9317-b27eba62a911"
```
WillEngler, line 24: Are all these little quote changes from a linter run or something?

OwenPriceSkelly: yeah black's default iirc is to never use single quotes unless replacing them risks changing the semantics

garden_ai/garden.py
OwenPriceSkelly: I did indeed lol

PR #60, opened 2023-02-15
46 garden create command

### LINKED ISSUE #29: write more readable json
# write more readable json

*As the author of a Garden, I'm habitually setting `year = "2022"` on
all of my gardens, even though it's now 2023. The generated
metadata.json should be as human-readable as possible, so that I (or
someone helping me debug) could spot this sort of mistake at a glance*

This is probably as easy as pretty-printing the json we write to the
metadata file, but there might be json fields which are likely to be of
particular human interest or which are especially typo-prone; we might
want to reorder those fields to appear towards the top of the file.

## Acceptance Criteria

- Given: a complete/valid but incorrect metadata.json generated for a
  Garden
- When: viewing that file
- Then: the contents should be as human-readable as possible while still being json

### LINKED ISSUE #46: Garden Create Command
_As a model contributor I want to be able to create a new garden so others can find my pipelines_

# Description
There is a command line interface to garden. It is installed with `pip install garden-ai` alongside the garden SDK.  It has a command line argument parser.

For this story we will add the `create` command to the CLI. This creates a new garden.
It takes command line arguments to pick up all of the mandatory garden properties. If command line args are missing it will prompt for values _wizard style_.

The result is a minimally valid `garden.json` with a UUID that was generated in the cli. The command will create a basic garden directory structure.

NOTE: the behavior implemented by the linked PR is slightly different from what's described here -- it scaffolds a directory for the user _if one is specified_, but if left blank it does not perform any scaffolding. Also a slight departure is the `garden.json`, which is written to `~/.garden/db/gardens/{garden uuid}.json` instead; if `--verbose` is set the contents of the json will be pretty-printed when created.

The basic garden directory structure will be:
```
{pwd}/
    models/
      .gitignore
    
    pipelines/
      {pipeline_name}.py  # not this story
    
    README.md # populated with provided description? 
    
    garden.json # generated from Garden object if user input is all valid

```
NOTE: best practice to do a `git init` for user or to ask them to `git init` before `garden create`? 

## separable tasks 
- [x] CLI/wizardry prompting part 
- [x] directory scaffolding part 
    - should be implemented as a function `f(directory : str) -> str` 
    - sets up the directory if the proposed name is valid (returning the input unmodified)
    - `raise typer.Exit(code=1)` otherwise
- [x] unit tests

### Tasks

- [x] garden init should check that the current directory is empty 
- [x] keep directory-making behavior analogous to `git init` 
- [ ] README with light documentation on the purpose of the CLI
- [x] PyPI packaging and deployment 
- [x] Decide on skeleton file structure
- [x] Create gh action
- [x] Wizard prompting
- [x] Tests for basic create action
- [x] typer/pydantic integration

- [x] `garden create` makes the skeleton directory (pathlib module has all the helpers we'd need)  
- [x] `garden create` prompts user and instantiates a Garden object and write garden.json 
...

# Assumptions:
1. The command is issued in the root folder of a newly created git repository.
2. The models directory is just a suggestion where users could put code for training models.
3. The models/.gitignore file has entries for common model files. We want to actively discourage users from saving their models in git. They should use the Garden model repository for this.
4. the pipelines directory should be empty 
5. the garden's metadata is stored locally at `~/.garden/db/{garden uuid}.json`, and can be printed to stdout by passing a `--verbose` flag

### LINKED ISSUE #56: split `models.py` into more focused modules
_`models.py` has become kinda unwieldy, and we still haven't thought of a better name for it. We're also reaching the point where we're talking about "pipelines" and "models" as distinct concepts (models must be used within a pipeline etc, any MLflow conversations)._

# Assumptions:
1. `models.py` -> `gardens.py`, `pipelines.py`, `steps.py`
2. `garden.py` -> `client.py` (?)
3. that joke about cache invalidation

# Acceptance Criteria
`models.py` no longer exists and/or is the future home of the model-specific features we'll support

OwenPriceSkelly:
closes #46, and bonus: closes #56, closes #29 

review note: if you want to try out the CLI, you should be able to do so by:
- checking out this branch locally
- `poetry install && poetry shell` 
- `garden` and `garden create` to heart's content
- (also please make a note of any colors that look bad/hard to read on your terminal so we can specify better defaults) 

----
:books: Documentation preview :books:: https://garden-ai--60.org.readthedocs.build/en/60/


garden_ai/app/create.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    Validate the directory provided by the user, scaffolding with "pipelines/" and
+    "models/" subdirectories if possible (i.e. directory does not yet exist or
+    exists but is empty).
+    """
+    if directory is None:
+        return None
+
+    if directory.exists():
+        if list(directory.iterdir()):
```
Aadit-Ambadkar, line 28:
```suggestion
        if len(list(directory.iterdir())) > 0:
```

garden_ai/app/create.py
```diff
@@ -0,0 +1,172 @@
...
+            logger.fatal("Directory must be empty if it already exists.")
+            raise typer.Exit(code=1)
```
Aadit-Ambadkar, line 30: rather than code=1, maybe something more descriptive would be better

garden_ai/app/create.py
```diff
@@ -0,0 +1,172 @@
...
+        authors = [name]
+        # prompt for additional authors until one is *not* given
+        while True:
+            name = validate_name(
+                Prompt.ask("Add another author? (leave blank to finish)")
+            )
+            if name:
+                authors += [name]
+            else:
+                break
```
Aadit-Ambadkar, line 133: this should be simplified to provide a list of comma-separated authors (in my opinion at least). It's less code and it's more clear.

garden_ai/app/create.py
```diff
@@ -0,0 +1,172 @@
...
+        if name:
+            contributors = [name]
+            while True:
+                name = validate_name(
+                    Prompt.ask("Add another contributor? (leave blank to finish)")
+                )
+                if name:
+                    authors += [name]
+                else:
+                    break
```
Aadit-Ambadkar, line 148: see above

garden_ai/app/create.py
OwenPriceSkelly: hmm yeah `list(...)` isn't the clearest truth condition, but I think the way I'd want to clean it up would be to replace it with something that plays nicely with iterators like `if any(directory.iterdir()): ...` rather than coercing to a list for the sake of `len`. Would also have the benefit of failing immediately in a nonempty directory rather than failing after `list`ing n directory items

OwenPriceSkelly: thanks for the attention to detail

garden_ai/app/create.py
OwenPriceSkelly: having a command return 1 is conventional to indicate failure to the parent process; iirc this comes from the C macros `EXIT_SUCCESS` and `EXIT_FAILURE` being usually defined as 0 and 1 on unix-likes, though I think technically anything nonzero is supposed to be interpreted as failure.

garden_ai/app/create.py
OwenPriceSkelly: I waffled on this too, but the problem is that people are likely to comma-separate their names as "Lastname, Firstname", so we'd need a different separator. My next thought would be semicolons, but that could be easy to forget or mistake in light of other conventions like `$PATH` being colon-separated etc. Figured that  `--author "lastname, firstname" --author "Familyname, Givenname"` would be more verbose but impossible to accidentally chop someone's name in half when parsing the input or something

OwenPriceSkelly: (though course I should be careful when I say "impossible" https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)

garden_ai/app/create.py
BenGalewsky: This "[truthy](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/)" test is more pythonic. So, I'd say keep it as is

BenGalewsky:
Now would be a good time to update the docs to show how to set up a garden

BenGalewsky:
Very nice PR! I love the commit history

OwenPriceSkelly:
^ thanks @BenGalewsky for the affirmation on the commit history! I took this as an opportunity to get the hang of re-ordering commits as well as squashing when I rebased, glad the effort shows :)

garden_ai/gardens.py
```diff
@@ -0,0 +1,246 @@
+from __future__ import annotations
```
WillEngler, line 1: [comment is not actually about this line of code]

A thought for follow-up work: I think we'll want more subdirectories under garden_ai. At least a directory for the pydantic models. It's awkward to have those at the same level as the autogenerated datacite file.

garden_ai/app/create.py
```diff
@@ -0,0 +1,171 @@
...
+        year=year,
+        description=description,
+        contributors=contributors,
+        tags=tags,
+    )
+
+    client.register_metadata(garden, out_dir=LOCAL_STORAGE / "gardens")
+
+    if verbose:
+        with open(LOCAL_STORAGE / "gardens" / f"{garden.garden_id}.json", "r") as f_in:
```
WillEngler, line 168: For follow-up work: I'd like to push us to just one local database.json file with all the state. Less futzing around with the filesystem, easier to whip up fixtures for testing.

docs/quickstart.rst
```diff
@@ -3,24 +3,93 @@ Quick Start
...
+    Usage: garden create [OPTIONS] [DIRECTORY]
+
+    Create a new Garden
+
+    ╭─ Arguments ───────────────────────────────────────────────────────────────────╮
+    │   directory      [DIRECTORY]  (Optional) if specified, this generates a       │
+    │                               directory with subfolders to help organize the  │
+    │                               new Garden. This is likely to be useful if you  │
+    │                               want to track your Garden/Pipeline development  │
+    │                               with GitHub.                                    │
```
WillEngler, line 30: Noting for follow-up work: Let's revisit this when we get to scaffolding pipelines. This help text suggests that invoking this directory option on garden creation is _the_ way to track your project in GitHub. This contradicts what I think will be our standard/encouraged path: maintaining your pipeline code in your pre-existing repo.

WillEngler (approved):
LGTM!

docs/quickstart.rst
OwenPriceSkelly: Ah I was trying to be as noncommittal as I could with the help text -- probably worth rephrasing before we demo but agree let's leave that as follow-up

garden_ai/gardens.py
```diff
@@ -0,0 +1,246 @@
...
+                                  for Big Pea Data extraction and classification.
+                                  It consists of a 2-hectare plot behind the monastery,
+                                  and a 30,000-plant dataset.'''
+
+        gc.register(pea_garden)
+
+    Notes
+    --------
+    Mendel's work was ignored by the scientific community during his lifetime,
+    presumably due to the lack of a working DOI.
```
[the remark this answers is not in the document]
OwenPriceSkelly: ^^ the real reason I do it

garden_ai/gardens.py
OwenPriceSkelly: yeah, I was noticing that in a big way looking at the generated documentation too, it was pretty ugly -- though on one hand, the `DataciteSchema` object and `Garden` object are both pydantic models, so it might be odd for only one of them to live in a `models/` subdirectory, idk. V{>0}'s problem.

PR #61, opened 2023-02-16
Add Validator decorator to _sync_author_metadata functions

### LINKED ISSUE #59: `_sync_author_metadata` whenever
> invoking `_sync_author_metadata` on pipeline or author validation is a good idea! It's also exactly the sort of thing we can't do if we're mutating the list in place (without assignment).

My proposed solution is to sync the metadata in any function which uses the author or contributor metadata. What are your thoughts on this?

Aadit-Ambadkar:
closes #59 

----
:books: Documentation preview :books:: https://garden-ai--61.org.readthedocs.build/en/61/

### DIFF
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -200,7 +200,8 @@ class Garden(BaseModel):
         except ValidationError as err:
             logger.error(err)
             raise
-
+    
+    @validator("pipelines")
     def _sync_author_metadata(self):
         """helper: authors and contributors of steps and Pipelines also appear as contributors in their respective Pipeline and Garden's metadata.
 
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -200,7 +200,7 @@ class Garden(BaseModel):
         except ValidationError as err:
             logger.error(err)
             raise
-    
+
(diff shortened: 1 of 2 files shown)


Aadit-Ambadkar:
@OwenPriceSkelly @blaiszik

OwenPriceSkelly:
hey @Aadit-Ambadkar - sorry to leave you hanging, it's been a busy week. A couple things:
1. the PR currently misses a quirk about pydantic validators, which is that they can only decorate class methods; I'm not familiar enough with how/why validators work like that under the hood to gauge how much work it would be to get around that restriction in this case
2. (More importantly) As of now, it wouldn't be worth your time to try and work around that: we're not yet settled on how much fine-grained author/contributor metadata users are even going to want, nor are we even settled on what should automatically count as a contribution 

We'll probably want to revisit this in the future, but for at least the next few sprints we're putting a pin in metadata issues in favor of building out the CLI and other moving parts; in the meantime I can keep an eye out for other issues that might make a good PR for you (if you'd like).

Aadit-Ambadkar:
hi owen. Sounds good. Please do keep an eye out for stuff I can work on. Thanks!

Issue #35, opened 2023-01-09
exhaustive unit tests for type hinting utils


OwenPriceSkelly:
*`utils.py` module currently contains `safe_compose` and an `issubtype` 
helper, the latter of which in particular should have exhaustive code 
coverage, due to the fact that `issubtype` will need to be able to handle
a slightly different flavor of type-hinting syntax for (apparently) every new
python release for the foreseeable future*

This story is intended to be primarily for writing unit tests and making
sure we're exhaustively handling all the possible syntaxes w/r/t python
version; several legal type-hint syntaxes are currently just ignored or
not yet handled at all (e.g. `Callable`, `Any`).

## Notes

Some examples of cases that may need to be handled or just disallowed
per-version (note also that many/all of these could be imported by someone using earlier versions via `typing-extensions` package):

- 3.8 introduces `Literal` types (unlikely to be useful, may not need
  special treatment)
- 3.9 introduces `Annotated` types (plausibly useful, not handled)
- 3.9 deprecates some typing special forms like `typing.List[<type>]` in
  favor of the `list[<type>]` syntax (useful, already handled)
- 3.9 introduces `Protocols` (unlikely in user code, not handled)
- 3.10 introduces `TypeAlias` (plausibly useful, may not need special
  treatment)
- 3.10 introduces `X | Y` syntax, conceptually equivalent to `Union[X,Y]`, but implemented via a
  `types.UnionType` unrelated to `typing.Union` (useful, already
  handled)
- 3.10 introduces `ParamSpec` and `Concatenate` (unlikely to show up in
  user code but maybe useful for us?)

And if we want to support 3.11:

- Variadic generics <https://peps.python.org/pep-0646/#abstract> seem
  likely to be useful in user code for things like annotating an array's
  shape – my guess is that this will be more common than doing so with
  `Annotated`
  - these would also greatly simplify a 3.11+ implementation of
    `safe_compose`

## Acceptance Criteria

unit tests covering the cases above for the appropriate python versions

PR #62, opened 2023-02-17
54 outsource issubtype

### LINKED ISSUE #28: disallow `Any` annotations
# disallow `Any` annotations

*as the author of a step, I might be tempted to label argument/return
types as the `typing.Any` type. This is not a good idea for steps, and I
should not be allowed to do that.*

The `Any` type only exists in the first place as a way to soften the
type-hint-adoption curve, which is counter to the point of our requiring
them for steps in the first place. Moreover, because the future of
python type-hinting is pretty shaky, we're likely going to need to
repair/reimplement our `issubtype` helper and limiting the set of
annotations we're willing to accept would simplify that in the future.

## Acceptance Criteria

- Given: `Any` in a step's annotation
- When: validating the step
- Then: user should be politely advised to do better

## Assumptions:

- if users can opt-out of annotation checking implicitly (by using `Any` for an annotation), they should also be allowed to do so explicitly (e.g. a `@step(check_type_hints=False, author=...)` kwarg to opt-out)
- implicitly/explicitly opting out should entail the same behavior under the hood

### LINKED ISSUE #32: use PEP 593 `Annotated[<type>, <info>]` to populate step `input_info`
_as a step author (using python 3.9+) I might have PEP 593-style annotations containing type metadata that I'd prefer not to lose. (if I went to the trouble of using the syntax, I probably did it on purpose)_

# Acceptance Criteria
Given a `step` with an `Annotated[<type>, <info>]` style argument- or return-annotation, when no other explicit `input_info=...`(and/or `output_info=...`) are given to the `step` decorator, then the `<info>` part of the `Annotated[<type>, <info>]` should be used as a default

### LINKED ISSUE #54: outsource `issubtype`
_due to python's "move gradually and break things" approach to type hint adoption, maintaining a consistently correct `issubtype` is going to be more work than it's worth (assuming it's even correct for all cases already, which it is not)._

# Assumptions:
1. We should replace it with `beartype.door.is_subhint` or similar, if any other battle-tested typing libraries have it implemented somewhere (though I could not find anything else promising)

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
this was a fun one -- closes #54, closes #32, closes #28 and obsoletes #35 thanks to the **very** thorough efforts of `beartype`: https://github.com/beartype/beartype#features. 

First class support for scientific-community-motivated 3rd party type hints like `numpy.typing` or typed PyTorch tensors also looks great. 

Incidentally, the beartype readme is also (if you don't mind their puns) imo the best resource I've stumbled across for making sense of the wide world of type-hint-PEPs (check out the "compliance" section)

----
:books: Documentation preview :books:: https://garden-ai--62.org.readthedocs.build/en/62/

### DIFF
--- a/garden_ai/utils.py
+++ b/garden_ai/utils.py
@@ -1,106 +1,14 @@
 import logging
 from inspect import Signature, signature
 from itertools import zip_longest
-from typing import Union
-import requests
 
+import beartype.door
+import requests
 from pydantic.json import pydantic_encoder
 from typing_extensions import get_args, get_origin
 
 logger = logging.getLogger()
-
-try:  # noqa C901 (mccabe exceeded due to try/except)
-    from types import UnionType  # type: ignore
-except ImportError:
-
-    def issubtype(a: type, b: type) -> bool:
-        """Helper: report whether `a` is a subtype of `b`.
-
-        This is defined slightly differently for versions which do not know about
-        `types.UnionType` (the `type(A | B)` special type)
-        """
-        if a == b:
-            return True
-        if a is None or b is None:
-            return False
-        # e.g. if a is `Union[x,y]`, then a_origin is `Union` and a_args is `(x, y)`
-        a_origin, a_args = get_origin(a), get_args(a)
-        b_origin, b_args = get_origin(b), get_args(b)
-        if a_args == b_args == ():
-            # case 1: no args
-            # (i.e. rules out anything fancy enough for square brackets)
-            return issubclass(a, b)
-
-        elif b_origin is Union:
-            # case 2: b is a union of types, which are the contents of b_args
-            if a_origin is Union:
-                # both are unions, so must have *all* the same args
-                return all(issubtype(x, y) for (x, y) in zip_longest(a_args, b_args))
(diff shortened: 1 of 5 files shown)


BenGalewsky (approved):
This is a great PR! Also, love the negative lines of code.

garden_ai/utils.py
```diff
@@ -1,106 +1,14 @@
...
-        elif issubclass(a_origin or a, b_origin or b):
-            # Case 3: other complex types, e.g. List[str] subtypes Sequence[str]
-            # the little `or`s are here because of the following fun fact:
-            # get_origin(tuple[x]) == get_origin(Tuple[x]) == get_origin(Tuple) == tuple
-            # ... get_origin(tuple) is None.
-            # I'm sure there's a very wise reason for this
-            return all(issubtype(x, y) for x, y in zip(a_args, b_args))
-
-        return False
+issubtype = beartype.door.is_subhint
```
WillEngler, line 11: what a satisfying diff in this file

WillEngler (approved):
LGTM!

PR #66, opened 2023-02-24
47 garden ai pipeline create

### LINKED ISSUE #47: Garden Create Pipeline
_As a model publisher I want to be able to add a pipeline to my ~garden~ life so I can ~thematically organize~ contain(?) my models_

# Description
For this story we will add the `garden-ai create pipeline` command to the CLI. This creates a new pipeline ~in the garden~.

It takes command line arguments to pick up all of the mandatory pipeline properties. If command line args are missing it will prompt for values wizard style.

The command generates a pipeline uuid and adds an entry to the `pipelines` list in `garden.json`

The command creates and populates a templated `{specified_or_current_directory}/{pipeline_name}/pipeline.py`  and empty package dependency files. 

# Assumptions:
1. could use a templating library like [jinja2](https://jinja.palletsprojects.com/en/3.1.x/)
2. `GardenClient.register_metadata` should read/write "local storage" which is a single file like `~/.garden/data/db.json`
    -  could be a good time to implement a little `show_gardens` / `show_pipelines` method for the client 
4. pipelines will need explicit authors rather than inferring them from enclosing garden (relax current default assumption of at least 1 garden per pipeline)
5. firm up CLI command word order (SVO vs SOV; verb-first like `garden create {garden, pipeline}` or object-first like `garden {garden, pipeline} create`? does `garden-ai garden {verb}` look less silly than `garden garden {verb}`?)
    - leaning towards noun-first, which is a slight refactor in the CLI but matches our sdk modules more closely, which seems convenient
7. templated file should contain instructive comments so users aren't surprised by e.g. type-hint annotations being required
8. there should be an entry in the local db (and so a UUID) for the pipeline before the templated file is created 
9. pipelines will need a valid python identifier, the CLI should validate this before creating a pipeline

OwenPriceSkelly:
closes #47 

----
:books: Documentation preview :books:: https://garden-ai--66.org.readthedocs.build/en/66/


templates/pipeline
```diff
@@ -0,0 +1,50 @@
+## THIS FILE WAS AUTOMATICALLY GENERATED {{  }} ##
```
WillEngler, line 1: Why do we have this file as well as garden_ai/templates/pipeline?

tests/test_client.py
```diff
@@ -110,3 +110,38 @@ def test_client_invalid_auth_token(mocker, mock_authorizer_tuple, token, mock_ke
...
     with pytest.raises(AuthException):
         GardenClient(auth_client=mock_auth_client)
+
+
+def test_local_storage_garden(mocker, garden_client, garden_all_fields, tmp_path):
+    # mock to replace "~/.garden/db"
+    mocker.patch("garden_ai.client.LOCAL_STORAGE", new=tmp_path)
+    uuid = garden_all_fields.uuid
+    garden_client.put_local(garden_all_fields)
+    record = garden_client.get_local(uuid)
```
WillEngler, line 120: It seems like the existing application code is still writing to {garden_id}.json right now, right? Is it a big lift to convert the code that's persisting models now so these parts aren't out of sync?

tests/test_client.py
```diff
@@ -110,3 +110,38 @@ def test_client_invalid_auth_token(mocker, mock_authorizer_tuple, token, mock_ke
...
+    uuid = pipeline_toy_example.uuid
+    garden_client.put_local(pipeline_toy_example)
+    record = garden_client.get_local(uuid)
+    assert record == pipeline_toy_example.json()
+
+
+def test_local_storage_keyerror(mocker, garden_client, garden_all_fields, tmp_path):
+    # mock to replace "~/.garden/db"
+    tmp_path.mkdir(parents=True, exist_ok=True)
+    mocker.patch("garden_ai.client.LOCAL_STORAGE", new=tmp_path)
```
WillEngler, line 136: Thinking out loud ... it would be great if our tests weren't doing a lot of filesystem mocking.  Probably not worth rethinking it on this PR, but I think we can get to a place where our tests are just manipulating a dict in memory and we can mock out the `read_from_disk` and `write_to_disk` calls.

WillEngler (commented):
Looking good! I have some feedback mostly on persistence

templates/pipeline
OwenPriceSkelly: we shouldn't, just slipped by me -- this is where I'd assumed jinja/poetry would look for templates but they needed to be down a level

garden_ai/app/pipeline.py
```diff
@@ -0,0 +1,207 @@
...
+from garden_ai import GardenClient, Pipeline, step
+from rich import print
+from rich.prompt import Prompt
+
+logger = logging.getLogger()
+
+pipeline_app = typer.Typer(name="pipeline", no_args_is_help=True)
+
+
+def validate_identifier(name: str) -> str:
```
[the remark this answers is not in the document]
OwenPriceSkelly: (for posterity after sync review) agreed, the word 'validate' was sticky in my brain because of the way pydantic "validators" work similarly (take a maybe-valid value and return a valid one) but that's very opinionated; even if we share pydantic's "parse, don't validate" mindset we don't need to share the "parse, don't validate; also you do the parse with something called `@validator`" mindset https://github.com/pydantic/pydantic/issues/578

garden_ai/app/pipeline.py
```diff
@@ -197,11 +197,11 @@ def dummy_step(arg: object) -> object:
         contents = template_pipeline(shortname, pipeline)
         with open(out_file, "w") as f:
             f.write(contents)
-        print(f"Wrote to {out_file}.n")
+        print(f"Wrote to {out_file}.")
 
-    client.put_local(pipeline)
     if verbose:
-        metadata = client.get_local(pipeline.uuid)
+        client.put_local_pipeline(pipeline)
```
WillEngler, line 203: I think we shouldn't store the pipeline before it's registered.

garden_ai/client.py
```diff
@@ -353,7 +359,41 @@ def register_metadata(self, garden: Garden, out_dir=None):
...
+                    data = json.loads(raw_data)
+
+        # update data['gardens'], leaving data['pipelines'] etc unmodified
+        gardens = data.get("gardens", {})
+        key, val = str(garden.uuid), garden.json()
+        gardens[key] = json.loads(val)
+        data["gardens"] = gardens
+        contents = json.dumps(data)
+
+        with open(LOCAL_STORAGE / "data.json", "w+") as f:
```
WillEngler, line 392: I think we're missing some opportunities to DRY our persistence operations. For example it looks like this block

```
 if (LOCAL_STORAGE / "data.json").exists():
            with open(LOCAL_STORAGE / "data.json", "r+") as f:
                raw_data = f.read()
                if raw_data:
                    data = json.loads(raw_data)
```

is present in put_local_garden and put_local_pipeline.

garden_ai/client.py
```diff
@@ -405,16 +448,50 @@ def put_local(self, obj: Union[Garden, Pipeline]) -> None:
...
-            data[key] = json.loads(val)
-            contents = json.dumps(data)
             f.write(contents)
         return
 
+    def get_local_pipeline(self, uuid: Union[UUID, str]) -> Optional[JSON]:
+        """Helper: fetch a pipeline record from 'local database', if one exists.
+
+        Find entry with key matching ``uuid`` and return the associated metadata
+        extracted from ``~/.garden/db/data.json``
```
WillEngler, line 466: I think the /db is a typo now

WillEngler (commented):
I think we should have one place where we read the data file and one place where we write to it.

PR #70, opened 2023-03-06
Upload model from CLI

### LINKED ISSUE #38: Publish SKLearn Model to Registry
_As a Garden User I want to publish my model to the registry so others can use it_

# Description
Users can upload their models to the Garden Model Registry with either the Garden CLI or with Garden SDK library calls. To keep this story simple, we will only work with SKLearn model flavors. Future stories will add additional model flavors.

Model names are namespaces with the user's email address.

Models can be uploaded with _extra pip requirements_ or _conda environments_

The command line version of this would be:
```shell
% garden-ai model register --extra-pip-requirements="pymatgen, matminer" --name matminer --flavor sklearn model.pkl
```

Model can be uploaded from the SDK with 
```python
garden.sklearn.log_model(sk_model, "matminer", 
                                             extra_pip_requirements=["pymatgen", "matminer"])
 ```

# Assumptions:
1. Tracking server URL can be set in the `GardenClient` constructor, but is defaulted to the same URL as the Garden backend
2. Passes the Globus Auth token to the server via the `MLFLOW_TRACKING_TOKEN` environment var
3. Implemented as a wrapper around MLFlow.sklearn.log_model function
4. Returns the version number for the published model

# Acceptance Criteria
Given I have an SKLearn model And I have created a garden client When I call `garden.sklearn.log_model` Then I can see the model listed in the Tracking Server UI

Given I have saved a pickled SKLean model When I issue the `garden-ai model publish` command Then I can see the model listed in the tracking server UI

### PR DESCRIPTION
**Pull request by WillEngler**
Builds on #68. Closes #38.

Adds a `garden model register` command that allows users to upload a model from the CLI.

## Limitations
- The backend is currently set up such that the client interacts with S3 directly. So only users who can read/write directly to the artifact bucket can use this. (So basically, only usable by Garden developers at the moment.)

----
:books: Documentation preview :books:: https://garden-ai--70.org.readthedocs.build/en/70/

### DIFF
--- a/garden_ai/app/model.py
+++ b/garden_ai/app/model.py
@@ -1,0 +1,58 @@
+from pathlib import Path
+from typing import Optional, List
+
+from garden_ai.client import GardenClient
+
+import typer
+import rich
+
+model_app = typer.Typer(name="model", no_args_is_help=True)
+
+
+@model_app.callback()
+def pipeline():
+    """
+    sub-commands for managing machine learning models
+    """
+    pass
+
+
+@model_app.command(no_args_is_help=True)
+def register(
+    name: str = typer.Argument(
+        None,
+        help=("The name of your model"),
+    ),
+    model_path: Path = typer.Argument(
+        None,
+        dir_okay=False,
+        file_okay=True,
+        writable=True,
+        readable=True,
+        resolve_path=True,
+        help=("The path to your model on your filesystem"),
+    ),
+    flavor: str = typer.Argument(
+        "sklearn",
+        help=(
+            "What ML library did you make the model with? "
+            "Currently we just support sklearn."
+        ),
+    ),
+    extra_pip_requirements: Optional[List[str]] = typer.Option(
+        None,
+        help=(
+            "Additonal package requirmeents. Add multiple like "
+            '--extra_pip_requirements "torch=1.3.1" --extra_pip_requirements "pandas<=1.5.0"'
+        ),
+    ),
+):
+    """Register a model in Garden. Outputs a full model identifier that you can reference in a Pipeline."""
+    if flavor != "sklearn":
+        rich.print("Sorry jk, we only support sklearn.")
+
+    client = GardenClient()
+    full_model_name = client.log_model(str(model_path), name, extra_pip_requirements)
+    rich.print(
+        f"Successfully uploaded your model! The full name to include in your pipeline is '{full_model_name}'"
+    )
--- a/garden_ai/app/model.py
+++ b/garden_ai/app/model.py
@@ -10,7 +10,7 @@ model_app = typer.Typer(name="model", no_args_is_help=True)
 
 
 @model_app.callback()
-def pipeline():
+def model():
     """
     sub-commands for managing machine learning models
     """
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by BenGalewsky** [comment]
> The backend is currently set up such that the client interacts with S3 directly.
Does this not go through the MLFlow server? Are we not proxying the S3 upload through it?

**Comment by WillEngler** [comment]
> Does this not go through the MLFlow server? 

By default the way we had it set up, no. I need to go configure the server with the `--artifacts-only` flag. That will proxy it through the server. I learned a lot in this issue thread. https://github.com/mlflow/mlflow/issues/629

File: garden_ai/app/model.py [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+from garden_ai.client import GardenClient
+
+import typer
+import rich
+
+model_app = typer.Typer(name="model", no_args_is_help=True)
+
+
+@model_app.callback()
+def pipeline():
```
OwenPriceSkelly, line 13: afaik it's ignored everywhere so this matters ~not at all, but could we rename the just-for-docstring function to `model` here

garden_ai/app/model.py
```diff
@@ -0,0 +1,58 @@
...
+        None,
+        help=(
+            "Additonal package requirmeents. Add multiple like "
+            '--extra_pip_requirements "torch=1.3.1" --extra_pip_requirements "pandas<=1.5.0"'
+        ),
+    ),
+):
+    """Register a model in Garden. Outputs a full model identifier that you can reference in a Pipeline."""
+    if flavor != "sklearn":
+        rich.print("Sorry jk, we only support sklearn.")
```
OwenPriceSkelly, line 52: lol

garden_ai/app/model.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    sub-commands for managing machine learning models
+    """
+    pass
+
+
+@model_app.command(no_args_is_help=True)
+def register(
+    name: str = typer.Argument(
+        None,
```
OwenPriceSkelly, line 23: heads up - typer reads `None` as "this is optional" and `...` as "required" when it sees either as a default https://typer.tiangolo.com/tutorial/arguments/optional/

garden_ai/app/model.py
```diff
@@ -0,0 +1,58 @@
...
+    ),
+    flavor: str = typer.Argument(
+        "sklearn",
+        help=(
+            "What ML library did you make the model with? "
+            "Currently we just support sklearn."
+        ),
+    ),
+    extra_pip_requirements: Optional[List[str]] = typer.Option(
+        None,
```
OwenPriceSkelly, line 43: we should probably add a short name like -x (just `Option(None, "--extra_pip_requirements", "-x", help="blah" )` instead

garden_ai/app/model.py
WillEngler: d'oh. Yeah this is some shameful copypasta

garden_ai/app/model.py
WillEngler: Ah, thanks! Definitely want it to be required

OwenPriceSkelly (approved):
lgtm!

PR #77, opened 2023-03-09
recreates mlflow download bug as integration test

OwenPriceSkelly:
This is the same bug that I was running into when trying to prep logan's real-world model for the 03/08/23 demo

This is a draft PR until the test passes 

to run just this test:
`$ pytest tests/integrations/test_mlflow.py` 

(note: this currently takes ~30-40 seconds on my laptop because it trains a tiny sklearn model every time)


OwenPriceSkelly:
see #80

Issue #65, opened 2023-02-23
Create Unified Pipeline Requirements.txt


BenGalewsky:
_As a pipeline developer I want to create a python virtual environment with all of the dependencies so I can debug my pipeline_

## Description
Needs to combine dependencies from the pipeline's model repositories along with the pipeline's local `requirements.txt`

Could we use [MLFlow's Virtualenv package](https://github.com/mlflow/mlflow/blob/master/mlflow/utils/virtualenv.py) to programmatically create pyenv  for a pipeline? 

We could do something similar to the `poetry shell` or `poetry run` commands to create a virtual env and use it.

```shell
% garden-ai pipeline shell rank-recipe 
```

This would consult the dependencies and create a pyenv (the MLFlow library computes a hash and re-uses a pyenv if it's been made before)

# Assumptions:
1. user has a `requirements.txt` file listing all the requirements needed for their pipeline
2. any models referenced in the pipeline have already been registered 
3. any models referenced in the pipeline appear as a default argument (so they can be inspected by the pipeline at init time, without needing to inspect the body of any steps) 

# Acceptance Criteria (edited)
- Given a pipeline instantiated with a path to a `requirements_file` ("./requirements.txt" by default in the template) and any underlying steps which look like `def f(..., Model('model-name')` (i.e. have a model download as a default arg), 
- when creating a `funcx.ContainerSpec` instance 
- then the pipeline correctly tracks conda and pip dependencies (i.e. `ContainerSpec(..., pip=pipeline._pip_dependencies, conda=pipeline._conda_dependencies)` would be correct)

OwenPriceSkelly:
looking ahead to #39, I think that we'll probably want the unified requirements.txt to be a conda `environment.yml` file, right?

WillEngler:
This seems like a @BenGalewsky question but I'll give my two cents to see if I understand the integration :)

From the ContainerSpec interface and also looking at the DLHub container service integration code, it seems like whether you're specifying packages to be installed via PyPI or Conda, you need a Python list at the end, not a flat file. ([ContainerSpec definition.](https://funcx.readthedocs.io/en/latest/reference/client.html#funcx.sdk.container_spec.ContainerSpec)) ([Example code of passing pip requirements to a ContainerSpec](https://github.com/DLHub-Argonne/dlhub_sdk/pull/184/files#diff-d9f342f8f4fc31b3d62c0f9faa6d22f1cff83b9789ef69c906263708d0eb86ffR27))

[...]

Issue #39, opened 2023-01-10
Build Container To Run Published Model


BenGalewsky:
_As a Garden user I want to be able to build a docker image so I can make inferences inside a funcX Worker_

# Assumptions:
1. Will use the funcX container service SDK
2. Library dependencies will be extracted from the model repository using the [model.pyfunc.get_model_dependencies](https://www.mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#mlflow.pyfunc.get_model_dependencies) interface along with the extra-pip-requirements for the pipeline.
3. Use the `conda` format and pass to the container service in the `conda` dependency list of the container spec
4. The publish operation will obtain the funcX container ID and store that in the database.json
# Acceptance Criteria
Given I have an SKLearn model with additional dependencies When I publish the model Then the container service will build an image that contains the SKlearn library and the additional dependencies And the garden's metadata record will contain the uuid of that container

PR #84, opened 2023-03-17
Unified pipeline requirements

OwenPriceSkelly:
Addresses #65, aiming at a smooth hand off to #39 

## Overview

This PR adds the following behavior to the initialization of pipeline/step objects:

steps now extract model-specific dependencies from any models referenced in a default argument, e.g. `@step`-decorated function like `def my_step(*arg, model=garden_ai.Model('...')): pass` will have recorded the dependencies listed in the mlflow model's conda.yml in `my_step.model_conda_dependencies` and/or `my_step.pip_dependencies`, per the `conda.yml`. Steps also track the python version for the model in `my_step.python_version`, if one is specified by the requirements file for the model.

pipelines collect any of their step-level dependencies in `pipeline.conda_dependencies` and/or or `pipeline.pip_dependencies`, as appropriate. Pipelines also include any pipeline-level dependencies found in `pipeline.requirements_file`, depending on the extension of `pipeline.requirements_file`. 

Pipelines also have a `pipeline.python_version` attribute for the container spec, which is either set explicitly in the constructor or inferred from `sys.version_info`. If versions conflict a warning is logged, but no exceptions are thrown.

## Discussion
- for #39 : 
  `ContainerSpec(..., pip=pipeline.pip_dependencies, conda=pipeline.conda_dependencies, python_version=pipeline.python_version)`

## Testing
unit tests included, plus some manual testing (and debugging) with Logan's real model with a real non-pypi dependency 

----
:books: Documentation preview :books:: https://garden-ai--84.org.readthedocs.build/en/84/

### DIFF
--- a/garden_ai/mlmodel.py
+++ b/garden_ai/mlmodel.py
@@ -74,10 +74,9 @@ def upload_model(
 def Model(model_full_name: str) -> mlflow.pyfunc.PyFuncModel:
     """Load a registered model from Garden-AI's (MLflow) tracking server.
 
-    Tip: for large models, using this as a "default argument" in a ``@step``-decorated
-    function will trigger the download as soon as the pipeline is initialized,
-    _before_ any calls to the pipeline. (Usage elsewhere should be fine, but
-    the pipeline won't be able to download the model until is actually called)
+    Tip: This is meant to be used as a "default argument" in a
+    ``@step``-decorated function. This allows the pipeline to download the model
+    as soon as the steps are initialized, _before_ any calls to the pipeline.
 
     Example:
     --------
--- a/garden_ai/mlmodel.py
+++ b/garden_ai/mlmodel.py
@@ -1,3 +1,4 @@
+import pathlib
 import pickle
 from functools import lru_cache
 from typing import List
@@ -105,4 +106,7 @@ def Model(model_full_name: str) -> mlflow.pyfunc.PyFuncModel:
     to use this function as a default value for some keyword argument.
     """
     model_uri = f"models:/{model_full_name}"
-    return load_model(model_uri=model_uri, suppress_warnings=False, dst_path=None)
+    local_store = pathlib.Path("~/.garden/mlflow").expanduser()
+    local_store.mkdir(parents=True, exist_ok=True)
+    # don't clutter the user's current directory with mystery mlflow directories
+    return load_model(model_uri=model_uri, suppress_warnings=True, dst_path=local_store)
--- a/garden_ai/pipelines.py
+++ b/garden_ai/pipelines.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import logging
+import pathlib
 from datetime import datetime
 from functools import reduce
 from inspect import signature
@@ -53,6 +54,7 @@ class Pipeline:
     version: str = "0.0.1"
     year: str = Field(default_factory=lambda: str(datetime.now().year))
     tags: List[str] = Field(default_factory=list, unique_items=True)
+    requirements_file: pathlib.Path
 
     def _composed_steps(*args, **kwargs):
         """ "This method intentionally left blank"
@@ -76,6 +78,13 @@ class Pipeline:
             logger.error(e)
             raise
         return steps
+
+    @validator("requirements_file")
+    def check_exists(cls, req_file):
+        req_file = pathlib.Path(req_file)
+        if not req_file.exists():
+            raise ValueError(f"Could not find {req_file}.")
+        return req_file
 
     def register(self):
         """register this Pipeline's complete Step composition as a funcx function
--- a/garden_ai/pipelines.py
+++ b/garden_ai/pipelines.py
@@ -143,7 +143,10 @@ class Pipeline:
         self.conda_dependencies = list(set(self.conda_dependencies))
         self.pip_dependencies = validate_pip_lines(list(set(self.pip_dependencies)))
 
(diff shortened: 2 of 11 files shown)


WillEngler (commented):
Wouldn't you know it, a Python dependency management task was a lot more nuanced than we expected it to be 🙃 Thanks for digging into the weeds on this.

The only potential showstopper is the behavior in `_collect_requirements`. I'm concerned about the fragility of some other pieces but it's possible that it's ok to get us to a working happy path.

garden_ai/pipelines.py
```diff
@@ -77,6 +82,49 @@ def check_steps_composable(cls, steps):
...
+
+        if not reqs.is_valid:
+            raise ValueError(f"Failed to read requirements file {req_file}.")
+
+        return req_file
+
+    def _collect_requirements(self):
+        """collect both conda and pip requirements to pass to funcx container service"""
+        for step in self.steps:
+            self.conda_dependencies += step.model_conda_dependencies
```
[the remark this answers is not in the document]
OwenPriceSkelly: nope -- I think I have this already fixed locally, just didn't stage everything I meant to in an earlier commit, sorry about that. [watch this space]

garden_ai/pipelines.py
```diff
@@ -77,6 +82,49 @@ def check_steps_composable(cls, steps):
...
+
+        if self.requirements_file.endswith((".yml", ".yaml")):
+            conda_deps, pip_deps = read_conda_deps(self.requirements_file)
```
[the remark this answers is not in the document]
OwenPriceSkelly: yeah agreed, felt kinda funny. could work around it, but I think it's just a quirk of using pydantic for validation -- validators are class methods that run before the actual instantiation of the object, so we call stuff like `self._collect_requirements` in ``__post_init_post_parse__``, which is always run before returning control to the `Pipeline(...)` call

garden_ai/utils.py
```diff
@@ -176,3 +179,38 @@ def extract_email_from_globus_jwt(jwt: str) -> str:
...
+
+    python_spec_re = re.compile(r"python[=<>! ]*")
+    pip_spec_re = re.compile(r"pip[=<>!]?")
+
+    # python_version = ".".join(map(str, sys.version_info[:3]))
+    python_version = None
+    conda_dependencies = []
+    pip_dependencies = []
+
+    for dependency in contents["dependencies"]:
```
WillEngler, line 201: Are being too trusting of .yml files we're being fed? If this yml file is not a compliant conda yaml, can we wrap this in some exception so that we're not just passing up a KeyError to the user on `contents["dependencies"]`?

garden_ai/pipelines.py
```diff
@@ -77,6 +84,70 @@ def check_steps_composable(cls, steps):
...
+        for step in self.steps:
+            self.conda_dependencies += step.conda_dependencies
+            self.pip_dependencies += step.pip_dependencies
+            py_versions[step.__name__] = step.python_version
+
+        self.python_version = py_versions["pipeline"] or py_versions["system"]
+        self.conda_dependencies = list(set(self.conda_dependencies))
+        self.pip_dependencies = list(set(self.pip_dependencies))
+
+        if len(set(py_versions[k] for k in py_versions if py_versions[k])) > 1:
```
WillEngler, line 141: This is a little confusing. I'd consider not inserting falsy values into py_versions.

garden_ai/pipelines.py
```diff
@@ -77,6 +84,70 @@ def check_steps_composable(cls, steps):
...
+            logger.warning(
+                "Found multiple python versions specified across this"
+                f"pipeline's dependencies: {py_versions}. {self.python_version} "
```
WillEngler, line 144: It seems like this warning would fire if the user's system python was different from the version they explicitly specified for their pipeline, even if the pipeline and its steps are all on the same page. Do we want that behavior?

garden_ai/pipelines.py
OwenPriceSkelly: I've pulled that expression out into a `distinct_py_versions` variable to make the condition more readable, but I do kinda want the error message to say `{'pipeline': None, 'my_preprocess': None, 'my_step_with_model': '3.8.6', ...}` so it's extra clear which steps are/aren't suspects when debugging. 

Especially since a common fix will probably be "set a known good version in the pipeline constructor", I wanted `'pipeline': None,` or `'pipeline': <version>` to show up in the error message in either case.

garden_ai/pipelines.py
OwenPriceSkelly: I have a feeling that more info is better here, especially since a 'system' version that isn't as expected could be causing problems with un-/pickling, or could be the case that it's been used to infer something by mlflow incorrectly. 

This is also why I opted to not raise any exceptions, just logging a warning instead

garden_ai/mlmodel.py
```diff
@@ -105,4 +105,7 @@ def run_inference(
     to use this function as a default value for some keyword argument.
     """
     model_uri = f"models:/{model_full_name}"
-    return load_model(model_uri=model_uri, suppress_warnings=False, dst_path=None)
+    local_store = pathlib.Path("~/.garden/mlflow").expanduser()
+    local_store.mkdir(parents=True, exist_ok=True)
+    # don't clutter the user's current directory with mystery mlflow directories
```
WillEngler, line 110: ❤️

WillEngler (approved):
LGTM!

PR #95, opened 2023-03-29
93 lazy download models

### LINKED ISSUE #93: Fix loading models too eagerly (i.e. before registering with funcx)
_The current implementation of `garden_ai.Model(...)` eagerly downloads the MLFlow model with a given URI, and python's "default arguments are evaluated once when defined" behavior gets us some caching behavior for free. The pre-downloaded object is also currently inspected by the `Step` decorating it, in order to extract model-specific dependencies_

However, because it's the default value for one of the function's arguments, the model is now considered part of the function's definition; when the function definition is serialized and registered with/ uploaded to funcx, _the entire `Model` object is included_ as part of the definition. 

Note that this only applies to `garden_ai.Model(...)` expressions that appear as default args -- a model that is only downloaded in the body of a function is not a problem here. 

tldr: steps can infer model-level dependencies at definition time, because the model is part of the definition (yay), but steps also load the entire model into memory at definition time, because it's part of the definition (boo)

# Assumptions:
1. the call to `garden_ai.Model(...)` should not download the entire model unnecessarily 
2. the step still needs to infer the model-specific dependencies correctly 
3. when registering a pipeline, the model should not be uploaded 

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
Fixes #93 
## Overview

- when we register with funcx and serialize the function object, we also serialize any objects included in the default argument list. 
- This meant that we were accidentally including the entire model implicitly (in the form of a large `mlflow.pyfunc.PyFuncModel` object) when we registered a pipeline containing it.
- Now, the object implicitly serialized (a new `garden_ai.mlmodel._Model` object) is *much* smaller because it waits to load an underlying `mlflow.pyfunc.PyFuncModel` until it's actually necessary to make a prediction -- i.e. the pipeline is actually running.

## Discussion

More details in #93 

## Testing

Just needed to update the fixtures included earlier 

----
:books: Documentation preview :books:: https://garden-ai--95.org.readthedocs.build/en/95/

### DIFF
--- a/garden_ai/mlmodel.py
+++ b/garden_ai/mlmodel.py
@@ -5,6 +5,7 @@ from typing import List
 
 import mlflow  # type: ignore
 from mlflow.pyfunc import load_model  # type: ignore
+from utils.misc import read_conda_deps
 
 
 class ModelUploadException(Exception):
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
Awesome!!

Issue #73, opened 2023-03-07
Support tensorflow and pytorch MLFlow model flavors


WillEngler:
_Short description_

Currently, `garden-ai model register` only works with scikit-learn models. But there are a lot of other model [flavors](https://mlflow.org/docs/latest/models.html#built-in-model-flavors) that we'd like to support. We'd like to start by adding pytorch and tensorflow. 

# Assumptions:
1. Users will bring their own ML libraries. It so happens that the garden PyPI package currently requires scikit-learn because it's a requirement of mlflow. But we probably should not make other ML libraries required for users to install the garden SDK.
2. We assume that users will train their models out of band of garden. They will have a serialized version of the model on their local filesystem in whatever the common serialization format for the library they're using.
3. It will be ok for some internal code organization to change. For example, the GardenClient's log_model method currently doesn't even take an argument for the kind of ML Flow flavor it's dealing with.

# Acceptance Criteria
Given an authenticated user and a pytorch model , when I run `garden-ai model register "model-name" "/model/path" "pytorch"`, then the model gets uploaded to Garden's ML Flow instance. Ditto for tensorflow.

WillEngler:
Closed by #100 and #101

PR #100, opened 2023-04-01
Tensor & PyTorch support

steve-barnard:
Related Issue: #73 

## Overview

This PR is to provide initial functionality to allow a successful upload of models built in the PyTorch or TensorFlow flavor.
Example CLI arguments:
 - `garden-ai model register Steves-testTF-model ../garden-flavor-test/savedtf/ tensorflow`
 - `garden-ai model register Steves-testPT-model ../garden-flavor-test/pytorchtest1.pt pytorch `

## Discussion

The methods employed include the usage of the base library (pytorch or tensorflow) to enable loading of the serialized object subsequent to using the matching flavor of ML Flow to then log the model to the garden MLFLow URI. 

This PR also has the following changes:
- added .vscode/settings.json to .gitignore
- garden_ai/app/model.py
    - added exists check for model_path typer arguments
    - added dir_okay for model_path typer arguments (needed for tensorflow)
    - updated help text for model path and extra_pip_requirements
    - updated the rich text to raise a typer.BadParameter to allow the CLI to exit rather than sit and require manual abort that existed with the rich text alone
    - added flavor as a parameter
- garden_ai/client.py
    - added flavor parameter to log_model method
- garden_ai/mlmodel.py
    - imported torch and tensorflow needed for loading serialized versions.
    - added code to ignore a cpu vs gpu informational warning from tensorflow; hence the flake ignore of E402 for the import of tensorflow after a line of code
    - added the flavor parameter to ModelUploadException class
    - updated docstring
    - created conditional upload based on model flavor, could look into switch case potentially in the future.
        - also checks that the model is in the proper file/dir required for the flavor
    - updated the `model_full_name` in the Model method to align with the other references in the code to model name so that all match to `full_model_name`.
- test/integrations/test_mlflow.py
    - created a fixture for pytorch toy model
    - created a test that follows the sklearn scaffolding for pytorch but seems to have an issue with downloading the model (to be investigated as noted below in the testing section)
    - if this testing issue for pytorch is rectified, then it should be easy to continue to add testing support for further flavors, but I most likely need assistance to confirm this PR is properly uploading and that there are no further errors introduced in the changes listed above.

## Testing

To test simply apply the same structure above (the CLI args) and have a model saved according to one of the flavors [tensorflow](https://www.tensorflow.org/tutorials/keras/save_and_load#save_the_entire_model) via model.save(model_path) (which requires a directory, unless an HDF5 file) or [pytorch](https://pytorch.org/docs/stable/generated/torch.save.html?highlight=save#torch.save) via torch.save(model_path).

At this time I have reverted the parametrized unit tests which created toy models for the tensorflow and pytorch models; however due to the added complexity and many variables that could go into said models when compared to a serialized toy sklearn model parameterized tests might not be the way to go here. I feel this warrants further discussion on how to best test the full functionality and if models require additional parameters such as signatures and other items identified in the mlflow.flavor.log_model() documentations. 

[pytorch - log_model() docs](https://mlflow.org/docs/latest/python_api/mlflow.tensorflow.html#mlflow.tensorflow.log_model) & [tensorflow - log_model() docs](https://mlflow.org/docs/latest/python_api/mlflow.tensorflow.html#mlflow.tensorflow.log_model)

For now, I have a test for the pytorch model that should work as a template, however there seemed to be issues I could not debug 
Exact error: `RuntimeError: PytorchStreamReader failed reading zip archive: invalid header or archive is corrupted`
(the pytorch model was failing at unzipping, this could be an issue with me attempting to download the Model [via Model(full_model_name)] and should be investigated further - perhaps a new issue created if this failure continues with other models/systems/users to rule out my exact circumstances) Steps to reproduce are to run the pytests for test_mlflow.py sklearn should pass and pytorch should fail.

----
:books: Documentation preview :books:: https://garden-ai--100.org.readthedocs.build/en/100/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,4 @@ dmypy.json
 /.dir-locals.el
 /.DS_Store
 /.github/.DS_Store
+.vscode/settings.json
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -16,7 +16,7 @@ If you have it, we recommend installing the CLI with `pipx`_.::
 
 You should then be able to see the following: ::
 
-    $ garden create --help
+    $ garden-ai garden create --help
 
     Usage: garden create [OPTIONS] [DIRECTORY]
 
@@ -73,7 +73,7 @@ which you can paste in your terminal window. This will generate credentials in
 your home directory, so you shouldn't need to repeat this step as the same user
 on the same computer. Here's what that might look like: ::
 
-    $ garden create --title=...  # etc
+    $ garden-ai garden create --title=...  # etc
 
     Authenticating with Globus in your default web browser:
 
--- a/garden_ai/app/model.py
+++ b/garden_ai/app/model.py
@@ -36,7 +36,7 @@ def register(
         "sklearn",
         help=(
             "What ML library did you make the model with? "
-            "Currently we just support sklearn."
+            "Currently we support the following flavors 'sklearn', 'tensorflow', and 'pytorch'."
         ),
     ),
     extra_pip_requirements: Optional[List[str]] = typer.Option(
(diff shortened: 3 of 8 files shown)


WillEngler:
This is fantastic. Thank you for the well-researched and tested contribution. I also appreciate the open questions and next steps in the PR description. Even with some of the snags you've described I'd like to err on the side of getting this merged in quickly to avoid drift and merge conflicts.

Right now I think the only "issue" is the torch and keras imports at module scope in some of the files. (Like [this](https://github.com/Garden-AI/garden/pull/100/files#diff-b745ba2b6f5b08e4c7b75ba670e27f1be6cb290b8dadf65175465faa4a3794e1R9)). For now we can just make torch and keras dependencies in pyproject.toml. But before declaring victory on V0 we'll probably want to go back and make them optional and only import them when a user is dealing with that specific library.

(It so happens that scikit-learn is a requirement of the MLFlow client so we've gotten away with assuming that all users have that installed. Hence the top-level `sklearn` imports. So we didn't have any pattern established for this and I didn't think to mention it when we were scoping the ticket, my bad.)

I know you're traveling for a while so I'll probably try and do the extra fiddly work to get this PR in this week. Thank you @steve-barnard!

WillEngler:
I made some little test and import tweaks on #101 and pulled this in. Thanks again Steve!

Yeah, for now just going with conditional imports based on the flavor argument. I think we'll probably want to add them as optional requirements in pyproject.toml for easy dev setup and to be able to run tests in CI that use them. But for now I just pulled in what you had.

Issue #40, opened 2023-01-10
Execute Registered Pipeline


BenGalewsky:
_As a Garden User I want to execute a funcx-registered Pipeline so I can execute it remotely, on a known endpoint, and obtain the correct output from local inputs_

# Description
Look up the funcX function ID from the database.json
Endpoint UUID will be defaulted. Baked into the client code.
Calls funcx.run
Block on the task completing

```python
wine_pipeline("All day rose", endpoint="123-45-56")
```

# Assumptions:
1. Container has been built and registered with funcX (#89)
2. Pipeline function has been registered with funcX (#51)
4. `Pipeline.__call__` will have a keyword argument for the endpoint uuid

# Acceptance Criteria
Given I have published a model When I invoke the pipeline (`__call__` magic method) Then a funcX task will be executed on the endpoint that returns my expected results

OwenPriceSkelly:
see: https://github.com/funcx-faas/funcX/pull/1083

PR #103, opened 2023-04-07
Monkeypatch FuncX client to get around clientside serialization bug

WillEngler:
Temporary expedient to unblock #40

## Overview

The FuncX client currently can't handle the way Garden composes functions together. (It effectively only serializes the top level function but leaves out the functions being composed.) This PR uses the idea @OwenPriceSkelly has proposed to the FuncX team as a proper fix.

## Discussion

Hopefully this will be around for a month tops until the fix gets into FuncX.

## Testing

I tested this on Owen's WIP remote execution branch. Without the monkeypatch I got the telltale KeyError at execution time. (See error report in Owen's PR: https://github.com/funcx-faas/funcX/pull/1083) With the monkeypatch included we get past that error and hit our pipeline code. 

----
:books: Documentation preview :books:: https://garden-ai--103.org.readthedocs.build/en/103/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -32,7 +32,7 @@ from garden_ai.mlflow_bandaid.binary_header_provider import (
     BinaryContentTypeHeaderProvider,
 )
 from mlflow.tracking.request_header.registry import _request_header_provider_registry  # type: ignore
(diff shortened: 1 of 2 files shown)


OwenPriceSkelly (approved):
lgtm! 🙈

PR #108, opened 2023-04-12
keep up with funcx -> globus-compute rebrand

OwenPriceSkelly:
## Overview

This looks big, but just renames all our references to funcx now that funcx is Globus-Compute™️

## Discussion

I erred on the side of renaming stuff like `funcx_*` to `compute_*` rather than `globus_compute_*`, so a few variable names might look a little odd. 

Also, the funcx rebrand changed `FuncXClient` to just `Client`; we might prefer to tweak those imports so that it reads `globus_compute_sdk.Client` in our codebase instead of `Client`.

## Testing

Made sure the usual tests passed after the refactor

----
:books: Documentation preview :books:: https://garden-ai--108.org.readthedocs.build/en/108/


garden_ai/client.py
```diff
@@ -1,14 +1,14 @@
+# mypy: disable-error-code="import"
```
OwenPriceSkelly, line 1: this disables mypy import checking for this file, because black wouldn't let me keep the `# type: ignore` comment on the same line as the `from mlflow.tracking.request_header.registry import ...`

garden_ai/client.py
```diff
@@ -111,22 +111,22 @@ def __init__(
...
         os.environ["MLFLOW_TRACKING_URI"] = GARDEN_ENDPOINT + "/mlflow"
         _request_header_provider_registry.register(BinaryContentTypeHeaderProvider)
 
-    def _make_funcx_client(self):
+    def _make_compute_client(self):
         scope_to_authorizer = {
             AuthScopes.openid: self.openid_authorizer,
             SearchScopes.all: self.search_authorizer,
-            FuncXClient.FUNCX_SCOPE: self.funcx_authorizer,
+            Client.FUNCX_SCOPE: self.compute_authorizer,
```
OwenPriceSkelly, line 126: The class attribute is still `FUNCX_SCOPE` as of 2.0.0

WillEngler (approved):
LGTM

PR #115, opened 2023-04-18
`RegisteredPipeline` class

### LINKED ISSUE #113: `RegisteredPipeline` -- class to support executing remotely from metadata only
_As a registered-pipeline-user, I should be able Just Doi It from a known doi/uuid, without needing import/instantiate the full pipeline object from actual code._

Problem: because plain `Pipeline`s wrap actual steps/functions and functions can't be json-serialized, a plain `Pipeline` can't be constructed from just its registered metadata. 

To solve this, we'll want a new `RegisteredPipeline` class which can be transparently de-/serialized from just the metadata we register (no `pipeline.py`!), which is still always enough to execute remotely with globus compute.

Plain `Pipeline`s will still exist in more or less their current state, but won't be able to execute remotely at all; in contrast `RegisteredPipeline`s won't be able to execute locally (because they're just metadata), only providing remote execution.

# Assumptions:
1. a `RegisteredPipeline` should be constructible from just a doi or uuid, corresponding to a known registered `Pipeline`. 
2. a `RegisteredPipeline` should only be able to execute remotely -- no local execution 
3. a plain Pipeline should only be able to execute locally

# Other notes/considerations
- db operations should be refactored where necessary s.t. they are in terms of `RegisteredPipeline`s only. 
- doi, etc can be optional for `Pipeline` now 
- double check affected tests

# Acceptance Criteria
## Remote execution 
Given: a previously-registered pipeline's DOI/uuid, and its corresponding `db.json` record 
When: initializing a "mere metadata" `RegisteredPipeline` from that record 
Then: I can execute the pipeline remotely without any access to the original pipeline object or code 
## JSON "just works":
contrast with plain `Pipeline`s:
```
In [8]: record = get_local_pipeline_by_uuid(plain_pipeline.uuid)

In [9]: Pipeline(**record)  # Breaks 
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[9], line 1
----> 1 Pipeline(**record)

File ~/Library/Caches/pypoetry/virtualenvs/garden-ai-u6Ad6Oqe-py3.10/lib/python3.10/site-packages/pydantic/dataclasses.py:325, in pydantic.dataclasses._add_pydantic_validation_attributes.new_init()

File ~/Library/Caches/pypoetry/virtualenvs/garden-ai-u6Ad6Oqe-py3.10/lib/python3.10/site-packages/pydantic/dataclasses.py:425, in pydantic.dataclasses._dataclass_validate_values()

ValidationError: 3 validation errors for Pipeline
steps -> 0 -> func
   ... # etc 

In [10]: RegisteredPipeline(**record)
Out[10]: RegisteredPipeline(...) # cool!
```

OwenPriceSkelly:
Closes #113 

## Overview

This PR introduces the `RegisteredPipeline` class, which is strictly metadata so that it can be de-/serialized without losing anything. 

This PR also changes the `Garden` class so that it only has a list of `pipeline_ids`, rather than full pipeline objects -- to access more than just the uuid of a garden's pipeline, the class now has a few helpers like `garden.collect_pipelines()` and `.expanded_metadata()`. 

## Discussion

All the database helper functions that previously returned a dictionary now return an object 
See also the description of #113 
## Testing

How did you test this? 
Painstakingly 

----
:books: Documentation preview :books:: https://garden-ai--115.org.readthedocs.build/en/115/

### DIFF
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
+import json
 import logging
 from datetime import datetime
-from typing import List, Optional, cast
+from typing import Any, Dict, List, Optional, cast
 from uuid import UUID, uuid4
 
-from pydantic import BaseModel, Field, ValidationError, root_validator, validator
+from pydantic import BaseModel, Field, ValidationError, validator
 
-from garden_ai.utils.misc import JSON
+from garden_ai.utils.misc import JSON, garden_json_encoder
 
 from .datacite import (
     Contributor,
@@ -21,6 +22,10 @@ from .datacite import (
 from .pipelines import RegisteredPipeline
 
 logger = logging.getLogger()
+
+
+class PipelineNotFoundException(KeyError):
+    """Exception raised when a Garden references an unknown pipeline uuid"""
 
 
 class Garden(BaseModel):
@@ -113,9 +118,7 @@ class Garden(BaseModel):
     tags: List[str] = Field(default_factory=list, unique_items=True)
     version: str = "0.0.1"
     uuid: UUID = Field(default_factory=uuid4, allow_mutation=False)
-    pipelines: List[RegisteredPipeline] = Field(
-        default_factory=list, include={"__all__": {"uuid"}}
-    )  # see: fetch_registered_pipeline below
+    pipeline_ids: List[UUID] = Field(default_factory=list)
 
     @validator("year")
     def valid_year(cls, year):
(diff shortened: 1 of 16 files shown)


garden_ai/gardens.py
```diff
@@ -100,37 +104,71 @@ class Config:
...
+        for uuid in self.pipeline_ids:
+            pipeline = get_local_pipeline_by_uuid(uuid)
+            if pipeline is None:
+                raise PipelineNotFoundException(
+                    f"Could not find registered pipeline with id {uuid}."
+                )
+            pipelines += [pipeline]
+        return pipelines
+
+    def expanded_metadata(self) -> Dict[str, Any]:
```
WillEngler, line 140: I like this approach!

garden_ai/globus_compute/remote_functions.py
```diff
@@ -15,8 +17,9 @@ def register_pipeline(
     container_uuid: str,
 ) -> str:
     try:
+        to_register = inject_env_kwarg(pipeline._composed_steps)
```
WillEngler, line 20: nit that probably makes sense to return to in the MLFlow auth PR: I think the name `inject_env_kwarg` makes it seem like the injection is happening right now, at registration time. It seems like it actually makes it possible to inject stuff at execution time. I'd try to think of a name that conveys that.

garden_ai/pipelines.py
```diff
@@ -161,7 +162,7 @@ def _collect_requirements(self):
 
         self.python_version = py_versions["pipeline"] or py_versions["system"]
         self.conda_dependencies = list(set(self.conda_dependencies))
-        self.pip_dependencies = validate_pip_lines(list(set(self.pip_dependencies)))
+        self.pip_dependencies = list(set(validate_pip_lines(self.pip_dependencies)))
```
WillEngler, line 165: Was something breaking or did you just notice dupes and wanted to be proactive?

WillEngler (approved):
Wonderful! This is some really thoughtful foundation work.

To be transparent, here was my review process on this one:
- Was anything obviously broken or sloppy? No.
- Do I think we hit on the exact right interface/philosophy for this stuff? Not sure yet but I think this is a really good step in the right direction. I'll play with this over the next week or so and then we can compare notes and adjust from there. But overall it seems like some good and thorough work.

garden_ai/pipelines.py
OwenPriceSkelly: yeah the latter, just noticed it was duplicating dependencies and wanted to be proactive

Issue #74, opened 2023-03-08
Inference on a Large Model


BenGalewsky:
_As a garden user I want to perform inference on a large model so I can get exciting scientific results_

# Assumptions:
1. These models don't go into the model repository
2. The inference step will have an endpoint uuid instead of a model ID from the repository (or field in the model metadata)
3. User has to create an MLModel directory structure in their environment. Use the standard MLFlow.predict
4. We write a custom flavor to allow `predict` on the custom model (assuming it's not just a big PyTorch model, but something specific)
5. We have the support of Eli's group to assist
6. This is easy to do. If we find lots of "gotcha's" then this becomes a spike story and provide more detailed stories of how to eventually implement this

# Description
Instead of downloading the model just call the function on the specified endpoint. We can assume that the model is mounted in the endpoint

# Acceptance Criteria
Documentation on setting up a custom funcx endpoint, how to create the MLModel db and creating a custom MLModel flavor.

A reference implementation of AlphaFold on Delta with an associated endpoint

BenGalewsky:
Actually even easier... it could be that the pipeline specifies the default FuncX Endpoint to run on and then the the inference step is nothing special. It can assume that the model lives in an expected place in the worker.

OwenPriceSkelly:
got sidetracked the other day poking around the mlflow codebase and started thinking about this story, just writing down what I think might be the easy way to do this so I don't forget: we could have an analogue to `garden_ai.Model("me@uni.edu-mylocalmodel/2")` called like `garden_ai.LocalModel(...)`, which just un-sets `os.environ['MLFLOW_TRACKING_URI']` before calling `load_model`. That way, if the endpoint's `./mlruns` local directory looks like mlflow expects, `LocalModel` would make mlflow "download" models from `./mlruns` instead of our backend; we can decide later how to handle the case where the endpoint doesn't have everything it needs in the `./mlruns` local directory (e.g. trying to run the pipeline on an endpoint which doesn't have the model saved locally)

OwenPriceSkelly:
while I'm at it: since this is for Eliu's AlphaFold model, we'll also probably need to implement a little alphafold mlflow model flavor (which will probably not be the hard part here)

WillEngler:
Closing as stale. We'll probably revisit this soon

PR #120, opened 2023-04-26
remote pipelines can set env vars

### LINKED ISSUE #102: make lazy-loaded models play nice with auth
_Now that `my_model=Model(...)` puts off the actual download until `my_model.predict` is called, the garden backend MLFlow registry needs to be accessible during the **execution** of the pipeline, not just the registration._

The environment variables `MLFLOW_TRACKING_{TOKEN, URI}` are currently set by instantiating a `GardenClient`, but in general that won't be possible on a remote endpoint. Moreover, we can't just inject these from the client instance used to register the pipeline, since a token which is valid at registration-time would just make the pipeline break as soon as it expired.

# Assumptions:
1. There exists a `RegisteredPipeline` class for remote execution (thanks, #113)
2. Calling a `RegisteredPipeline` can feed the pipeline a valid token etc for that execution by passing `_env_vars={'MY_ENV_VAR': 'env-var-value'}`, e.g. `registered_pipeline(*args, _env_vars={'MLFLOW_TRACKING_URI':'...', ...},**kwargs)` or by setting them on the instance like `registered_pipeline._env_vars={...}`
3. Setting the correct variables remotely will enable the lazy MLflow Model download, and maybe has implications for how we eventually implement #74 

# Acceptance Criteria
the `soup_edibility_predictor` pipeline ([gist](https://gist.github.com/OwenPriceSkelly/5ac5db81ee097fee0d426910e914acca)) should be able to register and execute on a funcx endpoint without failing the assertion that the model was successfully downloaded in the final step 

in general, we should be able to set environment variables per-execution for remote pipelines after this story and there should be integration tests to exercise that

OwenPriceSkelly:
Closes #102 

## Overview

- Fixes(?) code coverage 
- Remote execution via a `RegisteredPipeline`, when instantiated like `remote_pipeline = client.get_registered_pipeline(doi_or_uuid)`, will set the necessary env variables for the endpoint-side model download.
- `RegisteredPipeline`'s are callable, just like local ones, but also accept a `_env_vars` kwarg, e.g. `remote_pipeline(input_data, another_arg, _env_vars={"ENV_VAR":"ENV_VALUE"}, ...)` 
- It also has an attribute `remote_pipeline._env_vars`, which takes precedence over the kwarg if both are set. 

## Discussion

I ended up including the actual functional changes with #113's earlier PR, so it won't be obvious from this diff what is actually doing the work here -- check out the `utils.misc.inject_env_kwarg` helper and the tweak to how we register the pipeline with globus compute. 

The motivation for this was just to set the `MLFLOW_TRACKING_URI` and `MLFLOW_TRACKING_TOKEN` environment variables, but I was pleasantly surprised at how well the "funcx-middleware" approach worked; I think that if we need to implement anything more sophisticated for e.g. #74 this is probably a good way to do it. 

## Testing

New integration test(s) 👀 -- I also tried the same pipeline on the globus compute tutorial endpoint in lieu of the dlhub endpoint and the lazy model download worked successfully 🎉

----
:books: Documentation preview :books:: https://garden-ai--120.org.readthedocs.build/en/120/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -26,7 +26,7 @@ from rich.prompt import Prompt
 
 import garden_ai.funcx_bandaid.serialization_patch  # type: ignore # noqa: F401
 from garden_ai import local_data
-from garden_ai.gardens import Garden
+from garden_ai.gardens import Garden, PipelineNotFoundException
 from garden_ai.globus_compute.containers import build_container
 from garden_ai.globus_compute.login_manager import ComputeLoginManager
 from garden_ai.globus_compute.remote_functions import register_pipeline
@@ -357,6 +357,48 @@ class GardenClient:
         registered = RegisteredPipeline.from_pipeline(pipeline)
         local_data.put_local_pipeline(registered)
         return func_uuid
+
+    def get_registered_pipeline(
+        self, identifier: Union[UUID, str]
+    ) -> RegisteredPipeline:
+        """Return a callable ``RegisteredPipeline`` corresponding to the given uuid/doi.
+
+        Parameters
+        ----------
+        identifier : Union[UUID, str]
+            The previously registered pipeline's DOI or UUID. Raises an
+            exception if not found.
+
+        Returns
+        -------
+        RegisteredPipeline
+            Instance of ``RegisteredPipeline``, which can be run on
+            a specified remote Globus Compute endpoint (and knows how to
+            set the appropriate MLFlow environment variables).
+
+        Raises
+        ------
+        PipelineNotFoundException
+            Raised when no known pipeline exists with the given identifier.
+        """
+        is_doi = "/" in str(identifier)
+        if is_doi:
+            registered = local_data.get_local_pipeline_by_doi(str(identifier))
+        else:
+            registered = local_data.get_local_pipeline_by_uuid(identifier)
+
+        if registered is None:
+            raise PipelineNotFoundException(
+                f"Could not find any pipelines with identifier {identifier}."
+            )
+
+        self._set_up_mlflow_env()
+        registered._env_vars = {
+            "MLFLOW_TRACKING_TOKEN": os.environ["MLFLOW_TRACKING_TOKEN"],
+            "MLFLOW_TRACKING_URI": os.environ["MLFLOW_TRACKING_URI"],
+        }
+
+        return registered
 
     def publish_garden_metadata(self, garden: Garden):
         # Takes a garden, and publishes to the GARDEN_INDEX_UUID index.  Polls
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,14 +37,14 @@ optional=true
 [tool.poetry.group.test.dependencies]
 pytest = "^7.2.0"
 pytest-mock = "^3.10.0"
-coverage = "^6.5.0"
+pytest-cov = "^4.0.0"
 flake8 = "^5.0.4"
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -37,14 +37,15 @@ optional=true
...
-coverage = "^6.5.0"
+pytest-cov = "^4.0.0"
 flake8 = "^5.0.4"
 mypy = "^0.981"
 types-requests = "^2.20.0"
 safety = "^2.3.1"
 types-pyyaml = "^6.0.12.8"
 torch = { version = "^2.0.0"}
-tensorflow = { version = "^2.11.0", python = ">=3.8,<3.12" }
+# tensorflow = { version = "^2.11.0", python = ">=3.8,<3.12" }
```
WillEngler, line 47: Was this breaking something? (Maybe the >= 3.8 requirement in particular?) We'd like to be able to have tensorflow and torch available for tests

tests/integrations/test_pipeline_registration.py
```diff
@@ -1,13 +1,27 @@
...
 
 runner = CliRunner()
 
 
+@pytest.fixture(autouse=False)
+def do_not_set_mlflow_env_variables():
+    # actually, DO set mlflow env variables
+    # locally defining a fixture to override
+    # the one mocking `setup_mlflow_env`
+    pass
```
WillEngler, line 17: This feels kludgey (and was initiated by me, you're just seeing it to its logical conclusion). It's too bad that to set MLFlow variables, the function that needs the override has to use something called `do_not_set_mlflow_env_variables`. Some very quick Googling didn't turn up canonical ways to do setup/teardown for pytest outside of the fixture system so I don't really have an alternative to suggest atm. But I'd like us both to keep our eyes on this and get to something more intuitve.

tests/integrations/test_pipeline_registration.py
```diff
@@ -36,3 +50,50 @@ def test_run_registered_pipeline():
...
+    client.register_pipeline(pipeline, container_id)
+
+    # load from client, setting mlflow env vars
+    registered_pipeline = client.get_registered_pipeline(pipeline.uuid)
+
+    input_data = "i'm afraid of change"
+    # set arbitrary other env vars
+    test_var, test_val = (
+        "GARDEN_TEST_ENV_VARIABLE",
+        "... especially of a changing environment",
```
WillEngler, line 74: 👏

garden_ai/client.py
```diff
@@ -358,6 +358,48 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+        if registered is None:
+            raise PipelineNotFoundException(
+                f"Could not find any pipelines with identifier {identifier}."
+            )
+
+        self._set_up_mlflow_env()
+        registered._env_vars = {
+            "MLFLOW_TRACKING_TOKEN": os.environ["MLFLOW_TRACKING_TOKEN"],
+            "MLFLOW_TRACKING_URI": os.environ["MLFLOW_TRACKING_URI"],
+        }
```
WillEngler, line 399: The grand finale!!

WillEngler (approved):
The only non-nit I have is the comment about taking tensorflow out of the test requirements. Any way we can leave that in?

garden_ai/client.py
OwenPriceSkelly: 🎉🎉🎉

Issue #97, opened 2023-03-30
Add model metadata to database and enable linking Foundry data used for training


WillEngler:
We want to be able to link models to data they were trained on.

# Assumptions:
1. There will be a new top-level `models` object in the local data.json.
2. Each model has an optional linked_dataset field

# Acceptance Criteria
Given I search for a Garden that contains a pipeline that contains a model that was trained on a Foundry dataset, when I query it I see a link to the Foundry metadata in the returned record.

OwenPriceSkelly:
should we fold the `garden-ai model list` cli command into this story, since this is the story for the metadata that would be listed? or should those be different stories

WillEngler:
Probably time to incorporate Ben's connection schema: https://docs.google.com/document/d/1t0EeLSYB-UPaHs7zb8FZGYEqbfhNZXZB2iAewST3GwY/edit

WillEngler:
This ended up being pretty big so I needed to split it up. Phase 1 is up for review: https://github.com/Garden-AI/garden/pull/121

Also sorry I missed your question Owen ...

> should we fold the garden-ai model list cli command into this story, since this is the story for the metadata that would be listed? or should those be different stories

I'm fine either way, but I'd lean to not including it because this task is already pretty big.

WillEngler:
Also now that we have data in the "database" we can do something like `cat ~/.garden/data.json | jq ". | {models}"` as a workable model list command for now 😄

PR #121, opened 2023-04-28
Add model metadata to local DB on publication. Include option to link Foundry dataset.

WillEngler:
First phase of #97 

## Overview

Creates new top-level Model entity in the local "database". Saves model metadata to database during model registration. Also creates some new Pydantic models that make it easier to keep track of what data is going where:

- `LocalModel`: all the metadata we need from a user on their local model to register it.
- `RegisteredModel`: the metadata about a model that gets saved in the database and will get saved in the Search index.
- `DatasetConnection`: currently very specifically metadata about a Foundry dataset. Contained in a Connections list of a RegisteredModel as a first stab at implementing the [connection schema](https://docs.google.com/document/d/1t0EeLSYB-UPaHs7zb8FZGYEqbfhNZXZB2iAewST3GwY/edit).

This is only phase one. We still need to create an ownership relationship from RegisteredPipeline to RegisteredModel and then have that roll up into the denormalized Search representation of gardens. But this was plenty big for one PR, so this was a good place to stop and review.

## Discussion

Some refactoring in this PR:
- Split out the CLI tests into separate files.
-`upload_model` was getting to be a beast so I split it into sub-functions and DRYed part of it. (Might have to un-DRY it if different model flavors end up having quite different implementations.)

## Testing

Updated existing tests to cover newly added functionality like the option to link a dataset from the CLI. Added new test that exercises new DB functionality,

----
:books: Documentation preview :books:: https://garden-ai--121.org.readthedocs.build/en/121/


tests/cli/test_search.py
```diff
@@ -0,0 +1,61 @@
+from garden_ai.client import GardenClient
```
WillEngler, line 1: Nothing changed here - just split out

tests/cli/test_pipeline.py
```diff
@@ -0,0 +1,48 @@
+import string
```
WillEngler, line 1: Nothing changed here - just split out

tests/cli/test_garden.py
```diff
@@ -1,14 +1,8 @@
-import string
```
WillEngler, line 1: Nothing changed here - just split out

tests/cli/test_model.py
```diff
@@ -0,0 +1,43 @@
+import pytest
```
WillEngler, line 1: This is the CLI test file with actual changes! Do look at this one

WillEngler (commented):
Notes for reviewers

garden_ai/mlmodel.py
```diff
@@ -16,90 +17,150 @@ class ModelUploadException(Exception):
...
+
+    """
+
+    model_name: str = Field(...)
+    flavor: str = Field(...)
+    extra_pip_requirements: List[str] = Field(default_factory=list)
+    local_path: str = Field(...)
+    user_email: str = Field(...)
+    connections: List[DatasetConnection] = Field(default_factory=list)
+    namespaced_model_name: Optional[str]
```
OwenPriceSkelly, line 54: nit: while we're waiting for pydantic v2, looks like the preferred way to manipulate attributes on instances is to use `PrivateAttr` https://docs.pydantic.dev/latest/usage/models/#private-model-attributes

garden_ai/mlmodel.py
```diff
@@ -16,90 +17,150 @@ class ModelUploadException(Exception):
...
                 loaded_model = pickle.load(f)
-                mlflow.sklearn.log_model(
-                    loaded_model,
-                    user_email,
-                    registered_model_name=full_model_name,
-                    extra_pip_requirements=extra_pip_requirements,
-                )
-        elif flavor == "tensorflow" and pathlib.Path(model_path).is_dir:
+                log_model_variant = mlflow.sklearn.log_model
+        elif flavor == "tensorflow" and pathlib.Path(local_path).is_dir:
```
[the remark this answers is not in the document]
OwenPriceSkelly: looks like validators fire in order of appearance of their fields, and can include a `values` argument to get access to previously-validated fields -- [example](https://docs.pydantic.dev/latest/usage/validators/) from the docs

OwenPriceSkelly (changes requested):
just a couple of nits and some ✨pydantic pedantry✨

garden_ai/mlmodel.py
WillEngler: I don't want it to be private though. I just want it to be a function of other attributes.

garden_ai/mlmodel.py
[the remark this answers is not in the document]
OwenPriceSkelly: Oh no I agree, I just meant that a validator could enforce that the model_path string looks like a `directory/with/model` when flavor is pytorch vs looks like a `path/to/model.pkl` when flavor is sklearn, etc. That way, we'd only ever need `if flavor == "tensorflow": ...` instead of remembering to check both `if flavor == "tensorflow" and pathlib.Path(local_path).is_dir: ...`

garden_ai/mlmodel.py
OwenPriceSkelly: yeah no you're totally correct I misunderstood -- we do want this attribute in the json dump of the metadata, right? not sure why I thought we didn't when I wrote that

garden_ai/mlmodel.py
```diff
@@ -16,90 +17,150 @@ class ModelUploadException(Exception):
...
+    flavor (str): The framework used for this model. One of "sklearn", "tensorflow", or "torch".
+    extra_pip_requirements (List[str]), optional
+        A list of additional pip requirements needed to load and/or run the model.
+        Defaults to None.
+    local_path (str): Where the model is located on disk. Can be a file or a directory depending on the flavor.
+    user_email (str): The email address of the user uploading the model.
+
+    """
+
+    model_name: str = Field(...)
```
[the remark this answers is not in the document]
WillEngler: Yeah good idea

garden_ai/mlmodel.py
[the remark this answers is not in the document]
WillEngler: I think this is where I'm confused. As I understand it, `pathlib.Path(local_path).is_dir` is checking the filesystem. So you could do some kind of check in the validator for file endings maybe, but you would still need to check later that the file/directory is indeed a file/directory on the filesystem. I just don't see a lot of value in an upfront check that could only use heuristics.

garden_ai/mlmodel.py
WillEngler: > we do want this attribute in the json dump of the metadata, right?

Yeah exactly

garden_ai/mlmodel.py
```diff
@@ -16,90 +17,150 @@ class ModelUploadException(Exception):
...
+class RegisteredModel(BaseModel):
+    """
+    The ``RegisteredModel`` class represents all the metadata
+    we want to publicly expose about an ML model that has been registered with Garden.
+
+    Args:
+    model_name (str): A short and descriptive name of the model
+    version (str): What
+    flavor (str): The framework used for this model. One of "sklearn", "tensorflow", or "torch".
+    extra_pip_requirements (List[str]), optional
```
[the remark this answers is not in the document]
WillEngler: d'oh, fixed

OwenPriceSkelly (approved):
lgtm!

PR #126, opened 2023-05-04
Fetch remote garden from Globus Search

### LINKED ISSUE #123: Add ability to pull down Garden object from remote metadata
We need a way for SDK users to create a Garden object in their code by querying by DOI. This will allow users to execute pipelines remotely without having metadata about it saved locally.

# Assumptions:
1. There will be a new `get_garden_by_doi` method on the GardenClient object.
2. This will attempt to fetch a garden from Globus Search

# Acceptance Criteria
Given someone has published a garden, when I call my_client.get_garden_by_doi(published_doi), then I get a Garden object back populated with the remote garden's metadata.

WillEngler:
Closes #123 

## Overview

Adds `client.get_garden_by_id` and `client.get_garden_by_doi` methods that query Globus Search for gardens matching those IDs. It then constructs a Garden locally with the remote metadata and attaches the MLFlow variables needed to execute pipelines in the garden remotely.

Now we can do things like this with a Garden that has been published:

```
> from garden_ai import GardenClient
> gc = GardenClient()
> owen_garden = gc.get_garden_by_id("fcd7dca6-0622-4af5-a20d-6690418a1dae")
> import pandas as pd; test_df = pd.DataFrame({'smiles': ['c1ccccc1', 'n1ccccc1', 'o1cccc1']})
> res = owen_garden.ionization_potential_predictor(test_df, endpoint="86a47061-f3d9-44f0-90dc-56ddc642c000")
> print(res)
[7.84181567 8.49960351 7.8868247 ]
```

## Discussion

We need to give some of our Globus Search usage a second pass. For example, right now there is nothing stopping someone from posting lots of different gardens with the same DOI, which should probably be impossible.

Also, I did some refactoring of the search code. Basically I just pulled the stuff that hits Globus Search into a separate file.

## Testing

Added new unit tests to test how we handle different search responses and a new integration test for an end to end test of pulling a garden from remote metadata.


garden_ai/gardens.py
```diff
@@ -142,6 +142,24 @@ def pipeline_names(self) -> List[str]:
             names += [name]
         return names
 
+    def _set_pipelines_from_remote_metadata(self, pipeline_metadata: list[dict]):
```
OwenPriceSkelly, line 145: ok more qualitatively, this does make me a little nervous bc now there's a way to have a Garden instance refer to/call a pipeline even if it doesn't actually have that pipeline's ID registered -- I think the shortest way home is to just add a check to assert that `pipeline.uuid in self.pipeline_ids`, but I wonder if this is the "`RegisteredPipeline`" moment for gardens, i.e. one kind lives locally and can be messed with, and one is the stricter representation of a published record. Maybe `PublishedGarden` instantiates from search results and `Garden` instantiates from local data?

garden_ai/gardens.py
```diff
@@ -142,6 +142,24 @@ def pipeline_names(self) -> List[str]:
...
+        pipelines = []
+        for meta in pipeline_metadata:
+            try:
+                pipeline = RegisteredPipeline(**meta)
+                pipeline._env_vars = self._env_vars
+                pipelines.append(pipeline)
+            except ValidationError as e:
+                logger.warning(
+                    f"Could not parse pipeline: {json.dumps(meta)}. {e.__str__()}"
+                )
```
OwenPriceSkelly, line 159: if you add a bare `raise` here it'll also re-raise the validation error, which is usually pretty-printed/tries to be human readable

tests/integrations/test_search.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest.mark.integration
+def test_get_pipeline_from_gsearch():
+    client = GardenClient()
+    doi = "10.23677/w2ty-gv07"
+    garden_id = "fcd7dca6-0622-4af5-a20d-6690418a1dae"
+    garden = client.get_garden_by_id(garden_id)
+    same_garden = client.get_garden_by_doi(doi)
+    assert same_garden.pipeline_ids[0] == UUID("50039c98-b6c6-415a-b11b-9a0845c0a9b8")
+    assert garden.pipeline_ids[0] == UUID("50039c98-b6c6-415a-b11b-9a0845c0a9b8")
```
OwenPriceSkelly, line 15: I think `assert same_garden == garden` should work too here

OwenPriceSkelly: unless `@property` stuff makes that comparison weird

tests/test_search.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_get_by_doi_valid(mocker, valid_search_by_doi):
+    mock_search_client = mocker.MagicMock(SearchClient)
+    mock_search_client.search = mocker.Mock(
+        return_value=MockSearchResult(valid_search_by_doi)
+    )
+    garden = garden_search.get_remote_garden_by_doi("foo", {}, mock_search_client)
+    # Simple check that garden was populated with fields.
+    assert garden.year == "1863"  # Old timey Mendel year
```
OwenPriceSkelly, line 20: 👴🏻🫛

tests/integrations/test_search.py
WillEngler: Ah but the thing is that I don't want to just confirm they're the same. I want to see that it pulled down what I expected from the remote index. (Like they might equal each other but be wrong in the same way.)

garden_ai/client.py
```diff
@@ -388,6 +389,16 @@ def get_registered_pipeline(
...
+        self._set_up_mlflow_env()
+        return {
+            "MLFLOW_TRACKING_TOKEN": os.environ["MLFLOW_TRACKING_TOKEN"],
+            "MLFLOW_TRACKING_URI": os.environ["MLFLOW_TRACKING_URI"],
+        }
+
+    def get_email(self) -> str:
+        return local_data._get_user_email()
+
     def get_registered_garden(self, identifier: Union[UUID, str]) -> Garden:
```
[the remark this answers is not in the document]
WillEngler: I think I like the "both" idea!

OwenPriceSkelly (approved):
lgtm!

PR #146, opened 2023-05-25
garden create DOI print fix (#141)

### LINKED ISSUE #141: The CLI should always tell you the DOI of the resource it just created
When standing up a garden from scratch you need to make a pipeline, make a garden, and then link them together. Right now after creating a garden, the CLI doesn't tell you its ID. So to link the pipeline to the garden, you first need to go read through your data.json. Not fun.

Take care of this and any other cases like this.

# Assumptions:
1. Per #140, gardens will get a DOI upon creation.
2. When you create a garden (or do any operation on a garden) there will be some text from the CLI telling you the DOI of the garden that just got created.

# Acceptance Criteria
Given I am using the CLI when I create a garden, then the CLI prints out the DOI so I can use that in other commands.

MaxTuecke:
Fixes #141 

## Overview

Adds print of DOI on garden create. If no DOI exists, prints UUID.

## Testing

Added unit test to check if DOI is being printed to stdout on garden create

## Documentation

----
:books: Documentation preview :books:: https://garden-ai--146.org.readthedocs.build/en/146/

### DIFF
--- a/garden_ai/app/garden.py
+++ b/garden_ai/app/garden.py
@@ -141,6 +141,14 @@ def create(
     if verbose:
         metadata = json.dumps(local_data.get_local_garden_by_uuid(garden.uuid))
         rich.print_json(metadata)
+
+    if garden.doi:
+        rich.print(f"Garden '{garden.title}' created with DOI: {garden.doi}")
+    else:
+        rich.print(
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
Great turnaround on your first PR, no notes.

Issue #144, opened 2023-05-22
`steps` have their source code as metadata 


OwenPriceSkelly:
_As an intrepid user/colleague of garden users, I might find the source code of pipeline steps itself to be insightful/useful. Steps should have a metadata field containing the plain text snippet of their source code._

This could be a first step toward an eventual version-robust solution to #134 (and #132, probably) but is still valuable in the meantime.

# Assumptions:
1. wrapping something from `dill.source` module is probably the best bet
2. should be tested against steps defined in .py files, repl session, jupyter notebooks etc 

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
closed by #147

Issue #132, opened 2023-05-12
Feature request: Create/register a pipeline without making a separate text file. (Better notebook support)


WillEngler:
I was playing around with Ryan Jacobs's [MAST-ML notebook](https://colab.research.google.com/github/uw-cmg/MAST-ML/blob/master/examples/MASTML_Tutorial_7_ModelPredictions_with_CalibratedErrorBars_HostModelonFoundry.ipynb#scrollTo=TgecfqyYtVxF). At the end it shows you how to upload the models you've just trained to Foundry.

Garden's current pipeline registration approach assumes you are writing your pipeline in a separate Python file. This makes it awkward to develop a pipeline completely from within a notebook. Yes you can open a text editor in most notebook environments, but it's awkward that you can't make and test your steps inside of cells.

I don't think this is a must-have for V1, but it's useful to have in mind if/when we re-evaluate how we register functions on Globus Compute.

# Acceptance Criteria
Given I've created Garden "Step" functions inside a Jupyter notebook, I can use the Garden SDK to bundle them into a Pipeline and register the pipeline.

PR #147, opened 2023-05-27
144 steps copy source code

OwenPriceSkelly:
Completes #144 

## Overview

This adds a `source` field to steps, containing the snippet of source code users wrote defining the underlying function. This means that builtins and some other functions can't be used as steps, but they can still be easily used within steps, even if they're only one line. Also, we apparently already made it impossible to use builtins because they wouldn't have had annotations we could use -- I couldn't come up with any steps that would be invalid now but valid before this, anyway. 

```python
# also works as expected if this is in a notebook cell
@step
def preprocessing(input_data: pd.DataFrame) -> pd.DataFrame:
    """doing some preprocessing"""
    smiles = input_data["smiles"]
    return smiles
    
assert preprocessing.source == '@step\ndef preprocessing(input_data: pd.DataFrame) -> pd.DataFrame:\n    """doing some preprocessing"""\n    smiles = input_data["smiles"]\n    return smiles\n'
```

## Discussion

This isn't likely to add a ton of immediate value (this _could_ be useful for #132, depending on how we want to go about that), but regardless I think carrying a copy of the source code is a good-metadata-practice and I'd be surprised if we don't find ways to use it in the future, e.g. on the frontend I think it'll be nice eventually to be able to view the contents of steps. At the very least, if you're the sort of scientist to develop an otherwise lovely pipeline entirely in some `Untitled14.ipynb`, we're keeping a copy of the code you wrote safe somewhere you can't lose it. 

That being said, the `step.source` should not be treated as a source of truth -- there are many weird pathological edge cases that I don't _think_ we need to worry about, but make me nervous anyway. 

For example, if you (1) defined that `preprocessing` function in a plain python module (but not as a step); (2) imported it into a REPL/notebook; and then (3) edited the source file to read:
```python
# pipeline.py
def preprocessing(input_data: pd.DataFrame) -> pd.DataFrame:
    """doing some preprocessing"""
    frowns = input_data["frowns"]  # 😈
    return frowns
```
... and _then_ you constructed the step like: 
```python
# Untitled14.ipynb
my_step = Step(func=preprocessing)
```
You would have a step whose `func` is the code object corresponding to the original function, but whose `source` is the edited text which it only grabbed from the source file when it was asked to. 

I think I can still sleep at night, but only because I couldn't contrive an example that  used the `@step` decorator -- which makes sense, because you can only use the decorator syntax if you're defining the step at the same time you're defining the function itself. 

## Testing

I added a quick unit test and also sanity-checked in a couple notebooks, using steps defined in-notebook and imported from other modules

## Documentation

I updated the `Step` docstring and checked that it looked ok when I ran `mkdocs serve`, but reviewers could check the readthedocs PR build, too.

----
:books: Documentation preview :books:: https://garden-ai--147.org.readthedocs.build/en/147/

### DIFF
--- a/garden_ai/steps.py
+++ b/garden_ai/steps.py
@@ -8,7 +8,7 @@ from functools import update_wrapper, wraps
 from inspect import Parameter, Signature, signature
 from typing import Any, Callable, Dict, List, Optional
 
-from pydantic import Field, ValidationError, validator
+from pydantic import Field, validator
 from pydantic.dataclasses import dataclass
 from typing_extensions import get_type_hints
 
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
I appreciate the thorough discussion and context. I think this is a cool example of exploring a topic/concept and getting some working code out of it. Even if a frontend to display the source code isn't ready right away this is a good way to encode some discovery.

PR #148, opened 2023-05-30
Scaffolded pipeline exception handling in pipeline register

### LINKED ISSUE #130: Bug: Scaffolded pipelines are broken by default. (And the error message doesn't tell you why)
Steps to repro:
1. scaffold a new pipeline
2. try to register it

Expected behavior:
1. A no-op pipeline is registered

Observed behavior:
User gets an exception like " Could not parse /Users/will/Sandbox/seedlings/mastml/solute_energy_shift/pipeline.py as a Garden pipeline. Could not execute the Python code". This hides the more useful exception, "MlflowException: Not a proper models:/ URI: models:/YOUR MODEL's NAME HERE. Models URIs must be of the form ..."

Discussion:
It's debatable whether the scaffolded pipeline should be registerable as is. I think we should start users out with something that is known-working but maybe there's a case that we shouldn't. But we _definitely_ need to improve the error message.

MaxTuecke:
Fixes #130

## Overview

Adds better exception handling to pipeline register. Now shows user MLFlow errors during pipeline register and will explicitly warn the user when they try to register a scaffolded pipeline. 

## Testing

Adds two unit tests in test_filesystem_utils to check for appropriate errors in the cases of an invalid model name and trying to register a scaffolded pipeline. 

## Documentation

No documentation added.

----
:books: Documentation preview :books:: https://garden-ai--148.org.readthedocs.build/en/148/

### DIFF
--- a/garden_ai/app/model.py
+++ b/garden_ai/app/model.py
@@ -71,11 +71,6 @@ def register(
             f"Sorry, we only support 'sklearn', 'tensorflow', and 'pytorch'. The {flavor} flavor is not yet supported."
         )
 
-    if name == "YOUR MODEL's NAME HERE":
-        raise typer.BadParameter(
-            "Model name 'YOUR MODEL's NAME HERE' is reserved for scaffolded pipeline example. Please change your model name to something else."
-        )
-
     only_one_dataset_option_provided = (dataset_url and not dataset_doi) or (
         dataset_doi and not dataset_url
     )
--- a/garden_ai/app/model.py
+++ b/garden_ai/app/model.py
@@ -71,6 +71,11 @@ def register(
             f"Sorry, we only support 'sklearn', 'tensorflow', and 'pytorch'. The {flavor} flavor is not yet supported."
         )
 
+    if name == "YOUR MODEL's NAME HERE":
+        raise typer.BadParameter(
+            "Model name 'YOUR MODEL's NAME HERE' is reserved for scaffolded pipeline example. Please change your model name to something else."
+        )
+
     only_one_dataset_option_provided = (dataset_url and not dataset_doi) or (
         dataset_doi and not dataset_url
     )
--- a/garden_ai/mlmodel.py
+++ b/garden_ai/mlmodel.py
@@ -20,7 +20,7 @@ class ModelUploadException(Exception):
 
 
 class PipelineLoadScaffoldedException(Exception):
-    """Exception raised when a user attempts to load model with the scaffolded pipeline model name"""
+    """Exception raised when a user attempts to load model with the name SCAFFOLDED_MODEL_NAME"""
 
     pass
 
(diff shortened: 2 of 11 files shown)


garden_ai/utils/filesystem.py
```diff
@@ -31,6 +40,17 @@ def load_pipeline_from_python_file(python_file: Path) -> Pipeline:
     module = importlib.util.module_from_spec(spec)
     try:
         spec.loader.exec_module(module)
+    except ScaffoldedModelException as e:
+        raise ScaffoldedModelException(
+            "Failed to load model. It looks like you are trying to register a scaffolded pipeline."
+            + " Please replace YOUR MODEL's NAME HERE in your pipeline.py with the name of a registered Garden model."
```
WillEngler, line 46: Direct instruction on how to fix the issue right in the error message, perfect

garden_ai/mlmodel.py
```diff
@@ -193,6 +199,14 @@ def __init__(
         self.model = None
         self.model_full_name = model_full_name
         self.model_uri = f"models:/{model_full_name}"
+
+        # check if model name is 'YOUR MODEL's NAME HERE'
+        # lets users know if they are trying to upload a scaffolded pipeline.
+        if self.model_full_name == "YOUR MODEL's NAME HERE":
```
WillEngler, line 205: I think this is a good approach overall but I have a nitpick. This is a classic "magic string" situation where we probably want to pull "YOUR MODEL'S NAME HERE" into a constant that gets referenced both by the scaffolding code and the code that does the helpful check. That way if we change the name in one place it will still be right in all the other places. (Especially if we want to fix something subtle like that one weird lowercase 's'.)

garden_ai/app/model.py
```diff
@@ -71,6 +71,11 @@ def register(
             f"Sorry, we only support 'sklearn', 'tensorflow', and 'pytorch'. The {flavor} flavor is not yet supported."
         )
 
+    if name == "YOUR MODEL's NAME HERE":
+        raise typer.BadParameter(
+            "Model name 'YOUR MODEL's NAME HERE' is reserved for scaffolded pipeline example. Please change your model name to something else."
+        )
```
WillEngler, line 77: I'm not sure we need this check. I think we would already throw an error at Pydantic validation for `LocalModel` because model names aren't allowed to have spaces. (Or pretty much anything that isn't alphanumeric.) Also, while it's easy for someone to use the name in the scaffolded pipeline by mistake, someone would have to be deliberate to upload a model with this name. I'd lean towards taking out this check assuming that other validation would catch it, but open to other takes.

garden_ai/utils/filesystem.py
```diff
@@ -31,6 +40,17 @@ def load_pipeline_from_python_file(python_file: Path) -> Pipeline:
...
+            + "\nFor more information on how to use Garden, please read our docs: https://garden-ai.readthedocs.io/en/latest/"
+        ) from e
+    except MlflowException as e:
```
WillEngler, line 49: I think MLflow exceptions are broader than this. I think it's possible we could see an MLflowException because of, say, an auth error or a transient network error. I think you _could_ know that it's not the name of a registered model if it has a 404 error code.

WillEngler (commented):
Good approach overall! I have some nits and some design questions but this is mostly there.

garden_ai/templates/pipeline
```diff
@@ -39,7 +39,7 @@ def another_step(data: object) -> object:
 @step
 def run_inference(
     input_arg: object,
-    model=Model("YOUR MODEL's NAME HERE"),
+    model=Model("{{ scaffolded_model_name }}"),
```
WillEngler, line 42: Great!

WillEngler (approved):
Thanks for addressing my feedback! This looks good to go

PR #149, opened 2023-05-31
add naive approach to token refresh

### LINKED ISSUE #78: Make ML Flow connection attempts trigger token refreshes
Requests to Globus backend services made via the Globus SDK (like through the SearchClient) can detect when a user's credentials have expired and fetch an access token. But we don't go through Globus SDK calls when making proxied requests for MLFlow. We just put the garden access token into the MLFlow authorization header, and calls to MLFlow don't trigger a refresh. So it's possible for a user to make calls to MLFlow with an expired token that won't trigger a refresh.

# Assumptions:
1. 
2.

# Acceptance Criteria
Given I am a garden user who last went through the login flow three days ago, when I try to access MLFlow via the Garden client, then the client automatically refreshes my access token on my behalf and I can use MLFlow without getting a 403.

isaac-darling:
Fixes #78 

## Overview

Prevents users from needing to manually re-authorize by refreshing tokens as necessary before they are used. This is accomplished by ensuring they are valid before returning them in the new get method.

## Discussion

A quick run-through of the codebase revealed no further usages of the `client.garden_authorizer.access_token`, but any current and future uses of this property should instead be replaced with the new get method `_get_garden_access_token()`.

## Testing

This is a rather annoying problem to test, however, it could be done by simply waiting on an old token.

----
:books: Documentation preview :books:: https://garden-ai--149.org.readthedocs.build/en/149/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -40,7 +40,6 @@ from garden_ai.mlflow_bandaid.binary_header_provider import (
 from garden_ai.mlmodel import (
     LocalModel,
     RegisteredModel,
-    PossibleOldToken,
     upload_to_model_registry,
 )
 from garden_ai.pipelines import Pipeline, RegisteredPipeline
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
LGTM!

OwenPriceSkelly (approved):
lgtm!

PR #151, opened 2023-06-01
135 fix dependency conflict

### LINKED ISSUE #135: Bug: Can't specify ML library version that is different from the one in current garden environment
This is similar to #131 but has a different root cause.

I trained a Pytorch model in a Colab notebook using torch version 2.0.0. I downloaded the serialized model so I could register it from my laptop. I tried specifying `torch==2.0.0`. But this resulted in an error at container build time. 

```
Cannot install torch==2.0.0 and torch==2.0.1 because these package versions have conflicting dependencies. ERROR: ResolutionImpossible
```

When I did a `poetry install --all-extras` to get a Garden environment with torch in it, that grabbed torch 2.0.1. When the MLFlow client loaded the local model, it noted the version of torch that was installed locally and stored that in the requirements file in the MLModel directory. At pipeline registration time, the Garden CLI combined the requirements file I wrote myself with the model's requirements. That combined list has the contradictory requirements.

# Acceptance Criteria
Given I specify a version of a library in the requirements.txt of a pipeline, when I try to register the pipeline, then the container is told to use the version of the library that I specified.

OwenPriceSkelly:
fixes #135 

## Overview

Previously, there had been several places we collected dependencies that would eventually become part of the container spec:
1. from a model, whether implicitly via mlflow or explicitly via the CLI
2. from a step, whether explicitly by setting the attribute(s) or inferred from some model it uses (see 1)
3. the pipeline itself, explicitly via the attribute(s), explicitly via a requirements file, or implicitly from a step (see 2)

Now the only dependencies that the pipeline keeps are those explicitly set via the respective attributes or (typically) via a requirements file. 

Step- and model-level dependencies are still collected for the sake of warning users when there's a mismatch, but they don't have any effect on the actual packages requested in the container. 

## Testing

Updated old tests/fixtures to reflect the new desired behavior. 

## Documentation

Added a sentence to a note in the architecture overview clarifying that only explicit pipeline deps will go in the container.

----
:books: Documentation preview :books:: https://garden-ai--151.org.readthedocs.build/en/151/

### DIFF
--- a/docs/architecture_overview.md
+++ b/docs/architecture_overview.md
@@ -12,7 +12,7 @@ Broadly speaking, `step` and `Model` are lower-level abstractions for users to b
 A `Step` is the smallest "unit of code" usable to the Garden framework. A step is just a single function or callable that performs a specific task - such as pre-processing data or running inference - wrapped in some additional metadata (such as input/output type annotations) so that it can be composed with other steps in a Pipeline.
 
 > [!NOTE]
-> If your step references a `Model`, the step will attempt to collect its dependencies as part of its metadata.
+> If your step references a `Model`, the step will attempt to infer its dependencies as part of its metadata. These inferred dependencies won't be included in the container unless explicitly required by an enclosing `Pipeline`, but a warning will be logged for any apparent incompatibilities or mismatches.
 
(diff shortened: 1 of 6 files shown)


WillEngler (approved):
Nice! This will be really useful

PR #155, opened 2023-06-02
list and show cmds for gardens, models and pipelines.

### LINKED ISSUE #142: Make "list" CLI commands for all resource types
We have entries in our local datastore (json file) for gardens, pipelines, and models. As a user, once you've made a resource the only way to look them up later is to go and inspect your `data.json`. That is unpleasant. It would be good if we had a `list` command for each resource type that printed out the name/title and DOI (if available) of all the resources of that type you have. Bonus points for a `show` command that pretty prints more detailed information for a resource given its DOI.

# Acceptance Criteria
Given I am a CLI user who has made some Garden resources, when I issue the CLI command `garden-ai [resource type] list`, then I get a list of that resource from my data.json.

MaxTuecke:
Fixes #142

## Overview

Adds a list command for gardens, models and pipelines that displays a table of all the instances of that type of resource from the data.json file.

Adds a show command for gardens, models and pipelines that displays the json data of a specific resource from the data.json file.

## Testing

Added unit tests in test_garden for garden list and garden show
Added unit tests in test_model for model list and model show
Added unit tests in test_pipeline for pipeline list and pipeline show
Added unit tests in test_local_db for new helper functions _get_local_resource_json and _get_local_resource_table

## Documentation

No new documentation added.

### DIFF
--- a/garden_ai/local_data.py
+++ b/garden_ai/local_data.py
@@ -2,7 +2,7 @@ import json
 import logging
 from enum import Enum
 from pathlib import Path
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Union, List
 from uuid import UUID
 
 from garden_ai.gardens import Garden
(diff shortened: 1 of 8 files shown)


WillEngler (commented):
My big question is if we can get away without having to write the new "json" variants of the DB getters. Aside from that it's lower priority style/code organization concerns.

garden_ai/local_data.py
```diff
@@ -149,6 +150,164 @@ def _delete_old_model_versions(model_name: str):
...
+    -------
+    Table
+        Returns a rich table of all local models.
+    """
+    return _get_local_resource_table(
+        resource_type=ResourceType.MODEL, fields=fields, table_name=table_name
+    )
+
+
+def get_local_garden_json(garden_id: Union[UUID, str]) -> Optional[Dict]:
```
OwenPriceSkelly, line 263: (sorry for the wall of text) this is totally not obvious and something I probably could gone into earlier, but since this is kinda the reason we're using pydantic models/dataclasses for everything in the first place I'm jumping on the opportunity to elaborate:

This module shouldn't have public functions that put or get dicts/json directly. We want to stick to functions that put or get a complete instance of one of our dataclasses, so that other modules don't need to think about whether or not the json/dict they'd be using is valid. 

For example: consider the difference between the dict I'd get from your function, 
```
data1: Dict = get_local_garden_json("<uuid-bunchanumbers>")
``` 
vs from 
```
obj = get_local_garden_by_uuid("<uuid-bunchanumbers>")
data2: Dict = obj.dict()
```
The former would be a bit faster, because it returns the dictionary element directly instead of instantiating an object in the process. But what if the json it came from wasn't valid, e.g. the entry didn't have a `"doi"` key? 

In the latter case, whatever code uses that `data2` dictionary never needs to worry about whether it's a valid record -- without a `"doi"` key, `obj` could not have instantiated, so `data2['doi']` never needs to worry about KeyErrors, while `data1['doi']` does.

OwenPriceSkelly: that is, pydantic makes it natural to validate all your data before/when you instantiate (usually called parsing elsewhere, but pydantic calls them validators 🤷‍♂️)

tests/test_local_db.py
OwenPriceSkelly, line 1: tests look really thorough!

OwenPriceSkelly (changes requested):
This is a really good start, and the tests seem great. Some nits about local_data but those should be easy changes -- excited to see this merged!

garden_ai/app/console.py
```diff
@@ -1,3 +1,114 @@
 from rich.console import Console
+from rich.table import Table
+
+from typing import List, Optional, Any
 
 console = Console()
+
+
+def _get_rich_resource_table(
```
OwenPriceSkelly, line 9: Since this function is doing the "heavy lifting" in this module, could we clean it up a bit? I think something like this should be equivalent, and simplifying makes it easier to pick out the moving parts at a glance 
```suggestion
def _get_rich_resource_table(
    resource_objs: Optional[List[Any]], 
    resource_table_cols: List[str], 
    table_name: str
) -> Table:
    
    table = Table(title=table_name)
    for col in resource_table_cols:
        table.add_column(col)
    
    for resource_obj in resource_objs:
        row = []
        for field in resource_table_cols:
            row.append(str(getattr(resource_obj, field)))
        table.add_row(*row)
    
    return table
```

OwenPriceSkelly (changes requested):
This is looking good! Functionality/tests seem solid, just some cleanup nits to pick

tests/cli/test_console.py
```diff
@@ -0,0 +1,97 @@
...
+from typer.testing import CliRunner
+
+from garden_ai import local_data
+from garden_ai.app.console import (
+    get_local_garden_rich_table,
+    get_local_pipeline_rich_table,
+    get_local_model_rich_table,
+)
+
+runner = CliRunner()
```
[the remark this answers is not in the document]
MaxTuecke: Nope :)

OwenPriceSkelly (approved):
LGTM! nice work :)

PR #156, opened 2023-06-02
remove check prohibiting tf models saved as a single file

isaac-darling:
Remove check limiting `tensorflow` support to directories.

----
:books: Documentation preview :books:: https://garden-ai--156.org.readthedocs.build/en/156/


WillEngler:
Is it possible to generate a trivial small hdf5 example and add a test that hits this case? Kind of like the existing `test_mlflow_tensorflow_register`

tests/integrations/test_mlflow.py
```diff
@@ -159,3 +159,29 @@ def test_mlflow_tensorflow_register(tmp_path, toy_tensorflow_model):
...
     assert hasattr(downloaded_model, "predict")
+
+
+@pytest.mark.integration
+def test_mlflow_tensorflow_h5_register(tmp_path, toy_tensorflow_model):
+    # as if model.pkl already existed on disk
+
+    tmp_path.mkdir(exist_ok=True)
+    model_path = tmp_path / "tensorflowtest"
+    toy_tensorflow_model.save(model_path, save_format="h5")
```
WillEngler, line 170: If the only thing different about this test and the above test is `save_format="h5"`, try using `@pytest.mark.parametrize()` to factor this into a table-driven test: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

poetry.lock
```diff
@@ -1989,7 +1989,7 @@ cloudpickle = "<3"
 databricks-cli = ">=0.8.7,<1"
 docker = ">=4.0.0,<7"
 entrypoints = "<1"
-Flask = "<3"
+Flask = ">=2.2.5,<3"
```
WillEngler, line 1992: It looks like maybe you manually updated `poetry.lock`? If so I don't think that'll work. (It might get through this PR but the change will get blown away the next time this file is generated)

isaac-darling:
Main merged!

WillEngler (approved):
LGTM!

PR #176, opened 2023-06-12
Replace UUID with DOI

### LINKED ISSUE #140: Standardize DOI behavior and indexing
There are some oddities right now with how Garden handles IDs.
- DOIs should uniquely identify a resource of a given type, but there are multiple garden objects in the index with the same DOI.
- Some CLI commands take a UUID or a DOI, but some only take a UUID.

I propose the following constraints:
- In the search index and local DB, we key on DOI wherever possible. This will enforce unique DOIs for each Garden in the index.
- Only let users specify DOI when selecting something by ID in the CLI. This will remove confusion on which is the "real" id.
- Every time a user updates a pipeline or garden, we also update DataCite with the new metadata.

# Assumptions:
1. There is still a chicken/egg problem with pipeline scaffolding. We'll need to keep the auto-generated UUID in the scaffolded pipeline, but that is just something we'll use internally.

# Acceptance Criteria
Given I am a Garden user, when I perform some CLI operation where I need to specify a garden or pipeline, then I use its DOI and I am confident that will uniquely identify the garden or pipeline.

isaac-darling:
`Fixes #140`

## Overview

Remove UUIDs from Garden, and replace their role using DOIs.

## Discussion

There are substantial structural changes here, and `_mint_doi()` creates a chicken/egg problem.

## Testing

Tests had to be adjusted to conform to the new structure, so should be scrutinized in particular.

## Documentation

There were some naive documentation fixes, but not having read them all their could very well be more to be made.

----
:books: Documentation preview :books:: https://garden-ai--176.org.readthedocs.build/en/176/


garden_ai/client.py
```diff
@@ -253,13 +258,17 @@ def create_pipeline(
             data["authors"] = authors
         if title:
             data["title"] = title
+        data["doi"] = data.get("doi") or "placeholder"
 
         pipeline = Pipeline(**data)
-        record = local_data.get_local_pipeline_by_uuid(pipeline.uuid)
+        record = local_data.get_local_pipeline_by_doi(pipeline.doi)
```
OwenPriceSkelly, line 264: we won't need this check to reuse a DOI anymore, now that DOIs are the only thing we're indexing records on.

On that note, there's another higher-level design smell here that I think is worth pointing out -- if we're indexing our data on a single field, we really don't want to code around it in such a way that collisions are possible for that field, which is very much the case when introducing non-unique placeholder/default values for the index.

OwenPriceSkelly: My favorite example of this is the guy in California who registered the license plate "NULL", thinking he'd be impossible to ticket, but instead ended up getting hit every single time a traffic cop typo-d/skipped the license plate field writing the ticket. Here, it looks like the worst that happens is a redundant assignment, but we don't want our future selves to have to worry about this kind of thing if we can help it.

garden_ai/templates/pipeline
```diff
@@ -64,6 +64,5 @@ REQUIREMENTS_FILE = None  # to specify additional dependencies, replace `None`
     description="{{ pipeline.description }}",
     version="{{ pipeline.version }}",
     year={{ pipeline.year }},
-    tags={{ pipeline.tags }},
-    uuid="{{ pipeline.uuid }}",  # WARNING: DO NOT EDIT UUID
+    tags={{ pipeline.tags }}
```
OwenPriceSkelly, line 67: I feel like we may have touched on this synchronously with Will, but I _think_ we'll need the DOI as part of the scaffold (probably with the same warning comment) otherwise if registers a pipeline; edits it; re-registers the pipeline, we'll have minted (and permanently published w/ datacite) two different dois for the same entity.

tests/fixtures/search_results/valid_search_by_subject.json
```diff
@@ -1 +1,83 @@
...
+                                "input_info": "a spoonful of Soup object",
+                                "output_info": null
+                            }
+                        ],
+                        "contributors": [
+                            "Sister Constance",
+                            "Abbot Mortimer",
+                            "Friar Hugo"
+                        ],
+                        "doi": null
```
OwenPriceSkelly, line 75: let's make sure the "valid search result" has a DOI for the pipeline, too, even though we don't currently instantiate a pipeline straight from the search record

tests/integrations/test_datacite.py
OwenPriceSkelly, line 1: since integration tests are turned off by default, did you run pytest on these changed integration tests manually? `pytest -m "integration"` would do all integration tests and might be prohibitively slow, but you can pass pytest the specific file (something like `pytest tests/integrations/test_datacite.py`) and it'll just run the tests it finds there

OwenPriceSkelly (approved):
LGTM!

Issue #153, opened 2023-06-02
Make an automated Garden end to end test


WillEngler:
Garden integrates across a lot of different systems. To answer the question "is Garden working?" we can't just exercise our own code. But currently in CI runs we only execute unit tests and no integration or E2E tests.

Make a test that exercises the whole user-facing flow of Garden. Install it from PyPI, upload a model, register a pipeline, create and publish a garden, search for the garden, and remotely execute the pipeline via the garden. This should run on a [fixed schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) (like maybe daily) so that we can catch problems with changes to services we depend on. (As opposed to only running E2E tests on PRs.) Ideally we can also hook up the tests to run before/after cutting a new release.

# Assumptions:
1. There will be at least 2 ways to invoke the test. One pulls the latest Garden release from PyPI. Another uses a local copy of Garden so that devs can run it ad hoc.

# Acceptance Criteria
Given I am a Garden engineer, when I have just released or am about to release a new version of garden-ai, then there is a script I can run that will assure me that everything works or tell me what needs to be fixed.

WillEngler:
Example: https://github.com/MLMI2-CSSI/foundry/blob/main/tests/test_foundry.py

PR #178, opened 2023-06-15
Initializing a GardenClient with a Client Credentials Grant

MaxTuecke:
Part of (#153)

Currently when initializing a GardenClient, it always requires refresh tokens. However, if we want to initialize with a Client Credentials Grant, rather than the regular Authorization Code Grant, we won't have a refresh token.

## Overview

Changed GardenClient __init__ to also accept auth_client of type ConfidentialAppAuthClient to initialize with a Client Credentials Grant.

Added github secrets GARDEN_API_CLIENT_ID and GARDEN_API_CLIENT_SECRET to github action env vars.

## Discussion

In order to do end-to-end testing, we needed a Client Credentials Grant since the Authorization Code Grant requires the user to manually login.

## Testing

Added unit test for GardenClient init with Client Credentials Grant.

## Documentation

No documentation added.

----
:books: Documentation preview :books:: https://garden-ai--178.org.readthedocs.build/en/178/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -90,6 +90,9 @@ class GardenClient:
             "GARDEN_CLIENT_ID", "cf9f8938-fb72-439c-a70b-85addf1b8539"
         )
 
+        # If auth_client is type ConfidentialAppAuthClient, then we want to make our autherizers with
+        # ClientCredentialsAuthorizer to do a Client Credentials Grant. Otherwise we do an
+        # Authorization Code Grant and make RefreshTokenAuthorizer.
         if isinstance(auth_client, ConfidentialAppAuthClient):
             self.auth_client = auth_client
             self.openid_authorizer = ClientCredentialsAuthorizer(
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -79,7 +79,9 @@ class GardenClient:
     scopes = GardenScopes
 
     def __init__(
-        self, auth_client: AuthClient = None, search_client: SearchClient = None
+        self,
+        auth_client: Union[AuthClient, ConfidentialAppAuthClient] = None,
+        search_client: SearchClient = None,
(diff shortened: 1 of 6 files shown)


garden_ai/client.py
```diff
@@ -87,27 +92,56 @@ def __init__(
...
+            )
+            self.search_authorizer = ClientCredentialsAuthorizer(
+                self.auth_client, SearchClient.scopes.all
+            )
+            self.compute_authorizer = ClientCredentialsAuthorizer(
+                self.auth_client, Client.FUNCX_SCOPE
+            )
+            self.search_client = SearchClient(authorizer=self.search_authorizer)
+            self.garden_authorizer = ClientCredentialsAuthorizer(
+                self.auth_client, GardenClient.scopes.test_scope
```
WillEngler, line 143: Does this mean in GHA mode we're using the test_scope but in the "normal" mode we're still using the action_all scope? If possible I'd like these to be the same so that there aren't fiddly differences between the login methods. (So if possible just use the test scope for the first case too)

garden_ai/client.py
```diff
@@ -87,27 +92,56 @@ def __init__(
...
+            )
```
WillEngler, line 144: Currently the `_create_authorizer` method is responsible for stashing the user's email in the local datastore after the call to Globus Auth goes through. It seems like this isn't breaking anything yet, but I think you'll want to find a way to make sure that some email address is populated in the database after the ConfidentialAppAuthClient login.

WillEngler (commented):
Good stuff! My big concern is on us potentially requesting different scopes depending on the login method. Not sure if they even are different, but want to double check. If we're doing that and we can get the scopes requested to be identical I'd like that

WillEngler (approved):
Sweet, thanks for addressing the comments

Issue #166, opened 2023-06-09
Add upload and overwrite logic for presigned URL upload


WillEngler:
Make a new/altered implementation of model registration that uses the new presigned URL endpoint (https://github.com/Garden-AI/garden-backend/issues/18) to push a model directly to S3.

# Assumptions:
1. We're giving up the MLFlow artifact server's auto-versioning behavior for now. When a user uploads a model with the same name a second time, we will just overwrite it.
2.

# Acceptance Criteria
Given I am a CLI user, when I upload a model, then it is uploaded to S3 without going through AWS API Gateway.

PR #185, opened 2023-06-20
Will/direct s3 upload

WillEngler:
Implements #166

## Overview

There are two main parts to this ...

1. Add functionality for uploading model directories directly to S3 via presigned URLs. This is currently hidden behind a "feature flag" so that the old behavior is still the default until the download side is implemented. (Which I'll do next.)
2. This also dedupes logic around formatting headers and calling the Garden API by making a BackendClient. I just noticed we were repeating a lot of logic and wanted to consolidate and get closer to the pattern with our other external services.

## Testing

I manually tested all the backend calls I touched by making the BackendClient. (Search ingest, DOI creation, and presigned URL generation all work. DOI update is not working, but that's a preexisting problem.) Also manually tested model upload. Checked that output on S3 was as expected.

Also added a new test for the most involved part of the upload process: staging the local model directory.

## Documentation

No docs yet. No user-facing functionality changes in this PR.

----
:books: Documentation preview :books:: https://garden-ai--185.org.readthedocs.build/en/185/


garden_ai/client.py
```diff
@@ -295,9 +299,29 @@ def create_pipeline(
         return Pipeline(**data)
 
     def register_model(self, local_model: LocalModel) -> RegisteredModel:
-        registered_model = upload_to_model_registry(local_model)
-        local_data.put_local_model(registered_model)
-        return registered_model
+        if "S3_UPLOAD_FEATURE_FLAG" in os.environ:
```
WillEngler, line 302: This is where the upload feature is "hidden" for now. If you don't turn on this env variable, behavior should be exactly the same after this PR.

garden_ai/mlmodel.py
```diff
@@ -141,6 +145,74 @@ def upload_to_model_registry(local_model: LocalModel) -> RegisteredModel:
...
+            log_model_variant = (
+                mlflow.tensorflow.log_model
+            )  # TODO explore artifact path and sigs
+        elif flavor == ModelFlavor.PYTORCH.value and pathlib.Path(local_path).is_file:
+            import torch  # type: ignore
+
+            loaded_model = torch.load(local_path)
+            log_model_variant = mlflow.pytorch.log_model  # TODO explore signatures
+        else:
+            raise ModelUploadException(f"Unsupported model flavor {flavor}")
```
WillEngler, line 178: This whole first part of the function is copypasta from _push_model_to_registry. But in my next PR where I do the download side, I'll delete `_push_model_to_registry` so we'll be DRY again.

WillEngler (commented):
Notes for reviewers

garden_ai/client.py
```diff
@@ -295,9 +299,29 @@ def create_pipeline(
...
+                try:
+                    clear_mlflow_staging_directory()
+                except Exception as e:
+                    logger.error(
+                        "Could not clean up model staging directory. Check permissions on ~/.garden/mlflow"
+                    )
+                    logger.error("Original exception: " + str(e))
+                    # We can still proceed, there is just some cruft in the user's home directory.
+                    pass
+            registered_model = RegisteredModel(**local_model.dict(), version="1")
```
WillEngler, line 318: Another note for reviewers: I'm keeping the RegisteredModel class as is in this PR but I plan on changing it when I can delete the old code in the follow-up.

OwenPriceSkelly (changes requested):
I've got a couple gripes about raising/catching bare `Exception`s but other than that, this looks good to me!

garden_ai/constants.py
```diff
@@ -1,3 +1,13 @@
+import os
+
+_garden_dir = os.path.expanduser("~/.garden")
+
+
 class GardenConstants:
```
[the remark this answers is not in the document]
WillEngler: I think an Enum is clumsier because you would have to say GardenConstants.Garden_DIR.value. But I'd be open to just exposing the values directly. So in other files usage would look like ...

```
from garden_ai import garden_constants

print(garden_constants.GARDEN_DIR)
```

garden_ai/mlmodel.py
```diff
@@ -141,6 +145,74 @@ def upload_to_model_registry(local_model: LocalModel) -> RegisteredModel:
...
+
+        mlflow.set_tracking_uri("file://" + str(MODEL_STAGING_DIR))
+        experiment_id = mlflow.create_experiment("local")
+        with mlflow.start_run(None, experiment_id) as run:
```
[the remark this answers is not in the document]
WillEngler: Yeah basically that. I'll add a comment

OwenPriceSkelly (approved):
lgtm!

PR #187, opened 2023-06-22
fix datacite publishing bug

isaac-darling:
## Testing

I ran an end-to-end test successfully without error.

----
:books: Documentation preview :books:: https://garden-ai--187.org.readthedocs.build/en/187/


garden_ai/client.py
```diff
@@ -346,13 +346,24 @@ def _mint_draft_doi(self, test: bool = True) -> str:
...
+            titles=[{"title": metadata["title"]}],
+            publicationYear=metadata["year"],
+            publisher="Garden AI",
+            types={"resourceTypeGeneral": "Text"},
+            event="publish",
+            url=f"https://thegardens.ai/{obj.doi}",
+        )
+        del metadata["authors"]
+        del metadata["title"]
+        del metadata["year"]
```
WillEngler, line 365: The logic here looks good, thanks for figuring out what DataCite _actually_ wants.

Taking a closer look at this, and some of our existing code, could you implement this data transformation as `datacite_json` methods on the Pipeline and Garden classes? It looks like Owen had made some datacite-specific formatting logic that we would want to replace/amend if it isn't right. And that way we can add some easy unit tests that 
1. create a Garden and test that its datacite_json is as expected
2. create a Pipeline and test that its datacite_json is as expected

WillEngler: (And just in terms of old-school OOP fundamentals we keep the data transformation next to the data instead of mixing it in with this external API invocation method)

isaac-darling: I am not 100% sure I'm picking up what you're putting down, but I refactored the fix using that method and added a couple tests.

garden_ai/client.py
```diff
@@ -346,13 +346,20 @@ def _mint_draft_doi(self, test: bool = True) -> str:
...
+        metadata = json.loads(
+            obj.json() if isinstance(obj, Pipeline) else obj.expanded_json()
+        )
         metadata.update(event="publish", url=f"https://thegardens.ai/{obj.doi}")
+
+        metadata.update(json.loads(obj.datacite_json()))
+        if metadata.get("description"):
+            del metadata["description"]
+        del metadata["authors"]
+        del metadata["title"]
```
WillEngler, line 360: Can these last data munging steps live in the datacite_json methods?

tests/test_models.py
```diff
@@ -25,6 +26,29 @@ def test_validate_all_fields(garden_all_fields):
...
+
+    assert isinstance(data["creators"], list)
+    assert isinstance(data["titles"], list)
+    assert data["publisher"] == "thegardens.ai"
+
+
+def test_pipeline_datacite(garden_client):
+    @step
+    def four() -> int:
+        return 4
```
WillEngler, line 40: self-descriptive function names, great

WillEngler (approved):
Great! Thanks for grinding this out and going through all the revs

Issue #167, opened 2023-06-09
Add download and pipeline injection logic for S3 presigned URL approach


WillEngler:
The mirror of #166. Bundle in new logic for downloading the model into our Model class or similar. And instead of injecting environment variables for accessing the Garden API into pipelines at invocation time, bundle in presigned URLs for all models that the pipeline uses.

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a Garden CLI user, when I remotely execute a pipeline, then the remote pipeline can fetch any models it uses directly from S3.

PR #189, opened 2023-06-26
Download models directly with presigned S3 URLs

WillEngler:
Implements #167 

## Overview

Builds on #166 to fully implement direct S3 upload/download of models, bypassing our API Gateway for actually transferring the large model files.

**Breaking Changes**

- Because we're now pointing at a totally different backend model store, any models that have been made in Garden won't be able to be referenced going forward. It's a clean slate.

## Discussion

This PR includes some simplifications and removals. For example, it takes out step-level Python requirements. As a corollary, it also no longer gathers requirements from the MLModel itself.

The biggest 🤔 part of this PR is keeping environment variables as the way to smuggle in data to remote pipelines. It works well enough and there is already more than enough going on in this PR without trying to revamp that piece. But I feel like there is a better approach waiting to be found.

## Testing

Tested end to end on the DLHub endpoint by specifying "garden-ai@ git+https://github.com/Garden-AI/garden.git@will/s3-download" as my garden dependency. 

Notable omission in unit testing:
- The new methods that are "pure" file operations (`upload._upload_directory_to_s3_presigned` and `_Model.download_and_stage`) do not have unit test coverage. These are important functions, but because it's all external side effects, unit testing them would basically mean mocking every line and not be too useful. So the unit test coverage doesn't look great but these will get hit by e2e testing.

## Documentation

The PR updates some details and examples of models that are different now. (eg, a model's full name now looks like `willengler@uchicago.edu/my-model`)

----
:books: Documentation preview :books:: https://garden-ai--189.org.readthedocs.build/en/189/


garden_ai/client.py
```diff
@@ -436,18 +412,14 @@ def get_local_garden(self, doi: str) -> Garden:
...
             raise GardenNotFoundException(
                 f"Could not find any Gardens with identifier {doi}."
             )
-        self._set_up_mlflow_env()
-        registered._env_vars = {
-            "MLFLOW_TRACKING_TOKEN": os.environ["MLFLOW_TRACKING_TOKEN"],
-            "MLFLOW_TRACKING_URI": os.environ["MLFLOW_TRACKING_URI"],
-        }
-        return registered
+        self._generate_presigned_urls_for_garden(garden)
```
WillEngler, line 421: Note that we grab the URLs needed for all the pipelines in a garden when we make a garden from the client. And we make all the requests in sequence, not concurrently. So if someone makes a big garden with 10+ models this will be pokey. For now I just wanted to go for correct and easy to understand, but there are a lot of ways we could improve this and it could be a nice task for a student or outside collaborator.

WillEngler (commented):
Stray note for reviewers

garden_ai/steps.py
```diff
@@ -233,33 +226,3 @@ def wrapper(f):
             return Step(**data)
 
         return wrapper
-
-
-def inference_step(model_uri: str, **kwargs):
```
OwenPriceSkelly, line 238: surprised we still had this tbh

OwenPriceSkelly (approved):
some nits for you to pick before you merge, but nothing big enough to warrant the request changes button. lgtm!

garden_ai/mlmodel.py
```diff
@@ -215,88 +186,103 @@ def clear_mlflow_staging_directory():
...
+            local_model_path = self.download_and_stage(download_url, self.full_name)
+            self.model = load_model(local_model_path, suppress_warnings=True)
+            try:
+                self.clear_mlflow_staging_directory()
+            except Exception as e:
+                print(
+                    f"Could not clean up model staging directory. Check permissions on {self.staging_dir}"
+                )
+                print(str(e))
+                pass
```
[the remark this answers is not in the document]
WillEngler: I want to keep going here because this isn't a fatal error. There is just some cruft in the staging directory. I'd like to clean it up but the show doesn't have to grind to a halt.

Issue #169, opened 2023-06-12
Factor out Model class into skinny garden


WillEngler:
- We need Model download and execution in skinny garden
- Also, needs to tie in new S3 download logic. (Both URL injection and download staging in the remote execution environment)
- Also, can we consolidate down our Model classes? (There are currently 3. Could be 2?)

@OwenPriceSkelly to put meat on the bones.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
obsoleted by #190

Issue #168, opened 2023-06-12
Refactor steps and pipelines in anticipation of skinny garden


WillEngler:
- Simplify Step class
- Move validation from steps into pipelines
- Do the starter poetry stuff

Questions to answer:
- Will poetry work for the separate skinny garden package? (hopefully yes)
- Can we cut out Pydantic from the core skinny garden package? (hopefully yes)

@OwenPriceSkelly to fill in the ticket :)

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
some more story details as I'm thinking through it:
## Assumptions
1. steps (along with models) can move to a module with no intra-garden-sdk dependencies  
2. steps should not need much metadata beyond what is already "baked in" to functions, e.g. `__doc__`/`__name__`
	- pydantic potentially not a dependency for steps anymore? 
3. steps no longer fail to instantiate w/out annotations, but pipelines can fail w/ inconsistent annotations 
4. `mlmodel.py` can be moved to new package, but save refactoring for #169 to the extent possible 

## Acceptance criteria 
1. steps have been refactored to be dumber (enforced?)
2. pipelines are responsible for any "business logic" that currently lives in steps 
3. both questions answered 
	1. clear path to independent sub-package dependency with poetry or investigated alternatives
	2. nothing uploaded to pypi until #169
4. sanity-check a simple inference step in a clean environment with just skinny package installed locally 
5. tests for new/changed behavior 
6. updated docs

OwenPriceSkelly:
obsoleted by #190

PR #190, opened 2023-06-26
hack `_Model` so we don't need garden-core

### LINKED ISSUE #131: Bug: Pipeline can't use library versions that conflict with garden-ai dependencies
# Problem

I was trying to make a pipeline that uses [MAST-ML](https://github.com/uw-cmg/MAST-ML/tree/master). MAST-ML pins pyyaml == 5.4.1 in its requirements. `garden-ai` requires >= 6. Because we currently require garden-ai to be in the pip requirements sent to the container service, you can't include MAST-ML in a pipeline. The pyyaml versions conflict and registration fails at container build time.

There are tweaks we could do in garden or MAST-ML to get around this particular example. But this problem is more general. It will be unsustainable to keep garden-ai's requirements limited to the lowest common denominator of users' pipeline dependencies.

Imagine we want to include some common benchmarking or evaluation code at the garden level that does tabular data operations with Pandas. If we want to use pandas 2.x we will be preventing the registration of a ton of pipelines that use older pandas versions.

# Potential Approaches

In order from most kludgey to least ...

1. **Status Quo**. Try to limit the dependencies of garden-ai so that this doesn't hit _too_ many potential pipelines. Accept that users may not be able to use some libraries or some versions of libraries.
2. **Skinny Garden** . Similar to the `home_run` approach that DLHub took. Maybe we still need _something_ bundled in the container, but we can segment it into a separate package with a smaller number of dependencies so that conflicts are greatly reduced.
3. **No Dependencies**. We include neither a full nor skinny garden package with containers. We find a way to turn pipeline files into Globus Compute functions that have no dependencies on the Garden SDK. (Maybe they still depend on a skinny MLFlow client if we can't help it.)

# Acceptance Criteria
Given a user specifies a set of pip or conda dependencies that can be built by the container service, when they submit a pipeline with those requirements, then the container builds.

### LINKED ISSUE #133: Problems installing garden for development
While attempting to install and use `poetry` with garden, this error was encountered:

```
ImportError while loading conftest 'C:\Users\Isaac Darling\Desktop\Programs\globus\garden\tests\conftest.py'.
tests\conftest.py:9: in <module>
    import garden_ai
garden_ai\__init__.py:1: in <module>
    from ._version import __version__  # noqa  # type: ignore
garden_ai\_version.py:3: in <module>
    __version__ = importlib.metadata.version("garden-ai")
E   importlib.metadata.PackageNotFoundError: No package metadata was found for garden-ai
```

OwenPriceSkelly:
fixes #131 (!), obsoletes #168 and #169, also fixes #133 as far as I can tell.
## Overview

Garden induces no dependencies on remote endpoints besides mlflow-skinny (not even a `home_run`-style garden-core package!)

## Discussion

~Check out the comments in the `Model` function trying to explain the hack Still need to tweak any other code that is affected by this (like in steps/pipelines, plus tearing off the funcx bandaid) and come up with some tests, but putting this up here as a draft anyway for now.~

~Still WIP, but cleaned up and with all the moving parts in place. Some tests still need to be fixed and we have yet to replicate the demo with the new model downloading code (since that old model no longer exists 💔)~

The new ~hacks~ features live in `utils._meta`, which is also currently the only place they're documented. The big new moving part is `redef_in_main`, which re-execs an object's source code in the `__main__` namespace (unless its already there). We use this a few places:
- the `_Model` class gets redefined as `__main__._Model`. This happens whenever one is instantiated by the user (idempotent) so that the instance gets serialized without referencing `garden_ai`. 
- a new helper `garden_ai_compose_functions` will get `redef_in_main`d, too, in order to serialize a single dependency-free function with globus compute. We do this so that dill doesn't pick up a stray `garden_ai.utils.misc.safe_compose` namespace it won't be able to resolve. the inner function of this one is technically the function we *actually* give to funcx. 
- `_load_pipeline_from_python_file` is a similar but slightly different hack with comment-apologia for the interested 

This also tears off the funcx bandaid in favor of explicitly selecting our favorite serialization strategy (of the two)

This also moves the `_Model` class definition into its own module, with the idea being that the `garden_ai._Model` module has literally nothing but the singular class definition we're embedding in the function we serialize. There's no functional reason we need to do this, just wanted us to feel uneasy editing it in the future. 

## Testing

~Going to want to try and recreate the demo from 6/30 that downloads a model the new way, but proof-of-concept still seems sound.~ Successfully tested this end-to-end manually with the sklearn seedling! I also need to think harder about how we want to unit test this (if we even can), but since interfaces are staying basically the same I think we're ok leaning on the existing test suite. I also think that @MaxTuecke's incoming PR with the automated end to end tests should be enough of a canary in a coal mine to catch anything in the future that changes how serialization happens without us realizing. 

## Documentation

I've tried to be extra verbose with comments throughout, but there weren't many user-facing changes to document. Note to reviewers: please make me add more explanation wherever it's not yet clear, I've been steeped in this PR for a few weeks now and am sure there's more I could be explaining.

----
:books: Documentation preview :books:: https://garden-ai--190.org.readthedocs.build/en/190/


OwenPriceSkelly:
still TODO:
- safe_compose needs to be tweaked so it's consistent with the new `garden_ai_compose_functions` helper 
- stumbled on https://github.com/python/cpython/blob/e0d951d68aa360b30db0e85775814266e85272d2/Lib/runpy.py#L201, need to investigate further but at a glance we might be able to outsource one of the `exec`s to a builtin module, which if so wouldn't actually make us more secure but it does feel morally better 
- change prints to exceptions in `_Model` so that they're propagated back from funcx 
- think hard about if/how to unit test any of this

garden_ai/utils/_meta.py
```diff
@@ -0,0 +1,144 @@
+import inspect
+import linecache
+import textwrap
+
+
+def redef_in_main(cls):
+    """Helper: redefine a class in __main__, e.g. garden_ai._Model -> __main__._Model.
```
WillEngler, line 7: nit: This isn't _just_ for classes, right? Seeing as we apply it to `garden_ai_compose_functions`.

garden_ai/utils/_meta.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    def compose(*functions):
+        func, *funcs = functions
+
+        def inner(*args, **kwargs):
+            result = func(*args, **kwargs)
+            for f in funcs:
+                result = f(result)
+            return result
```
WillEngler, line 60: I'm pretty into this. You can have as many args in the "entrypoint" to your pipeline, but the rest of the steps can take and return one thing. That seems easy to explain.

garden_ai/utils/_meta.py
```diff
@@ -0,0 +1,144 @@
...
+    # console).
+    code_str = f"""
+class _USER_PIPELINE_MODULE:
+    {textwrap.indent(pipeline_code, '    ')}
+    """
+
+    # run the user's code as `__main__._USER_PIPELINE_MODULE` namespace
+    local_namespace: dict = {}
+    exec_getsource(code_str, __main__.__dict__, local_namespace)
+    # exec(code_str, {}, local_namespace)  # TODO see if this could work
```
WillEngler, line 135: Is this cruft now or do we still want to try this later?

garden_ai/model_file_transfer/upload.py
```diff
@@ -30,13 +28,15 @@ def _upload_directory_to_s3_presigned(
...
-        )
+    zip_filepath = shutil.make_archive("model", "zip", local_directory)
+    try:
+        with open(zip_filepath, "rb") as f:
+            files = {"file": ("model.zip", f)}
+            http_response = requests.post(
+                presigned_url.url, data=presigned_url.fields, files=files
+            )
+    finally:
+        os.remove(zip_filepath)
```
WillEngler, line 39: ahhhh, thanks for cleaning that up

garden_ai/pipelines.py
```diff
@@ -113,7 +112,7 @@ class Pipeline:
     tags: List[str] = Field(default_factory=list, unique_items=True)
     requirements_file: Optional[str] = Field(None)
     python_version: Optional[str] = Field(None)
-    pip_dependencies: List[str] = Field(default=[f"garden-ai=={__version__}"])
+    pip_dependencies: List[str] = Field(default_factory=list)
```
WillEngler, line 115: woooooooooooooooo

garden_ai/pipelines.py
```diff
@@ -195,6 +194,16 @@ def pip_deps_parsable(cls, pip_dep):
...
 
+    @validator("pip_dependencies", each_item=False)
+    def ensure_minimal_dependencies(cls, pip_deps):
+        import mlflow  # type: ignore
+
+        if not any(req.startswith("mlflow") for req in pip_deps):
+            pip_deps += [f"mlflow-skinny=={mlflow.__version__}"]
+        if not any(req.startswith("pandas") for req in pip_deps):
+            pip_deps += ["pandas<3"]
+        return pip_deps
```
WillEngler, line 205: Nice!

garden_ai/mlmodel.py
```diff
@@ -186,129 +183,6 @@ def clear_mlflow_staging_directory():
...
-        Returns
-        --------
-        Results of model prediction
-
-        """
-        self._lazy_load_model()
-        return self.model.predict(data)
-
-
 @lru_cache
```
WillEngler, line 186: Is this doing anything meaningful for us if the expensive operation is happening in `_Model`?

garden_ai/utils/_meta.py
```diff
@@ -0,0 +1,144 @@
...
+    # robust importlib equivalent goes too far for our needs -- we want `import
+    # garden_ai` in the user's code to refer to the already-imported garden_ai,
+    # but something about `exec_module`'s internals does a "true" re-import of
+    # garden_ai, with the unfortunate side effect of causing dill to think it
+    # needs to trace/serialize far more than it actually should.  (i.e.
+    # basically everything, including unserializable objects like the CLI's rich
+    # console).
+    code_str = f"""
+class _USER_PIPELINE_MODULE:
+    {textwrap.indent(pipeline_code, '    ')}
```
WillEngler, line 129: Does this make the very first line double indented? (because there are 4 spaces before the {} block)

Like it's clear that this works because you've tested this. Just a little confused

garden_ai/utils/misc.py
```diff
@@ -72,31 +44,7 @@ def safe_compose(f, g):
...
-            raise TypeError(
-                (
-                    f"Could not compose step {f.__name__} with step {g.__name__} "
-                    "due to return type signature mismatch: "
-                    f"expected tuple[{f_in}], got {g_out}."
-                )
-            )
-    elif len(f_in) == 1:
-        # case 2: return is a single value; verify that it's the only one
-        # expected by f.
```
WillEngler, line 99: love to see some red

WillEngler (approved):
Looks good! I just have some notes on docs/comments that you can address if it makes sense.

I spent some good heads down time reading and understanding. I want to do another pass to make sure I really grok it but all in all I think you did a good job with commenting, variable names, etc to make this reasonable to follow.

garden_ai/utils/_meta.py
```diff
@@ -0,0 +1,144 @@
...
+        import __main__
+
+        s = inspect.getsource(cls)
+        exec(s, __main__.__dict__)
+
+
+def make_func_to_serialize(pipeline):
+    """
+    Ensure that the composed function we register with globus compute is
+    "mainified", so that it has no references to the `garden_ai` namespace.
```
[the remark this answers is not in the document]
OwenPriceSkelly: 🦞🦞🦞

garden_ai/utils/_meta.py
OwenPriceSkelly: oh yeah good catch, that comment / the variable `cls` is outdated from when we were only using it for a class definition

garden_ai/mlmodel.py
OwenPriceSkelly: oh yeah you're totally right, actually think that hasn't been doing anything for us for a while now

garden_ai/utils/_meta.py
OwenPriceSkelly: oh great catch, I'm realizing that everything I tested it against started with an `#!/usr/bin/env python3` so it didn't cause problems but it easily could, I'll play around with it to sanity check

PR #191, opened 2023-06-28
Remove Funcx Bandaid

### LINKED ISSUE #188: BUG: import naming error
Installing garden-ai on a clean environment installs globus_compute_sdk 2.2.1. However, garden-ai depends on globus_compute_sdk 2.1.0. This causes a pretty significant problem:

In globus_compute_sdk 2.1.0, the Serializer Error class is:
```
class SerializerError:
    def __init__(self, reason):
        self.reason = reason

    def __repr__(self):
        return f"Serialization failed due to {self.reason}"
```

In globus_compute_sdk 2.2.1, the Serializer Error class is:
```
class SerializationError(Exception):
    def __init__(self, reason):
        self.reason = reason

    def __repr__(self):
        return f"Serialization failed due to {self.reason}"
```

See the git blame here: https://github.com/funcx-faas/funcX/commit/821234d1bb39fdc73975d01ffda18e3bf1daac32

The small difference between `SerializerError` and `SerializationError` causes an import error in the funcx bandaid which prevents the cli from working entirely.

2 possible fixes:
1. require globus_compute_sdk to be 2.1.0
2. require globus_compute_sdk to be 2.2.0 or greater, and change Serializer to Serialization

I am in favor of option 2, because it looks like the change will stay permanently

Aadit-Ambadkar:
Fixes #188 

Removes Funcx Bandaid (No longer needed)

----
:books: Documentation preview :books:: https://garden-ai--191.org.readthedocs.build/en/191/


Aadit-Ambadkar:
@WillEngler do you know why these are breaking? They're getting AuthAPIError (401) when trying to connect with globus sdk

WillEngler:
Hey @Aadit-Ambadkar. I don't know why that test would be failing. I made a fresh branch to see if I could reproduce the error without making any changes and I couldn't.

In any event, we can't remove the bandaid without specifying the the right serializer to use. I think @OwenPriceSkelly is on this.

PR #192, opened 2023-06-29
remove extra "model.zip" during upload

isaac-darling:
## Overview

Fixes a bug that caused model uploading to fail due to a path suffix being appended in two places.

## Testing

I ran an end-to-end test that went all the way through with no errors.

----
:books: Documentation preview :books:: https://garden-ai--192.org.readthedocs.build/en/192/


WillEngler (approved):
Sweet, thanks for jumping on this!

PR #195, opened 2023-07-03
[test, do not merge] Make a trivial change to test remote unit test execution

WillEngler:
Aadit is seeing a unit test failure where it seems like a test is failing on a globus_sdk.services.auth.errors.AuthAPIError. Should a unit test even be reaching out to the Globus backend this way? Why wasn't it triggered before? I'm seeing if I can repro.

----
:books: Documentation preview :books:: https://garden-ai--195.org.readthedocs.build/en/195/

### DIFF
--- a/garden_ai/app/garden.py
+++ b/garden_ai/app/garden.py
@@ -250,7 +250,7 @@ def add_pipeline(
         ),
     ),
 ):
-    """Add a registered pipeline to a garden"""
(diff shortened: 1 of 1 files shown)


OwenPriceSkelly:
I think I may have just been running into the same thing after bumping our globus dependencies -- a change in the globus sdk expected more attributes from a mock object in one of our tests than we had bothered to define, should be fixed now in #190

PR #198, opened 2023-07-06
add batching to presigned url requests

### LINKED ISSUE #193: Improve load time of gardens with many models
As of the new direct S3 download changes, gardens request presigned URLs for all the models in their pipelines at client.get_published_garden time. They request these URLs one by one. This could be made faster by batching. (Send one request to the backend for all the presigned URLs.) There might be other approaches too.

isaac-darling:
`Fixes #193 `

## Overview

Switches the request structure during loading to employ batching to reduce http overhead.

## Discussion

Requires the corresponding [backend PR](https://github.com/Garden-AI/garden-backend/pull/35).

## Testing

Cannot be fully tested until the aforementioned backend PR is merged.

----
:books: Documentation preview :books:: https://garden-ai--198.org.readthedocs.build/en/198/


isaac-darling:
As expected, `mypy` is not a fan of this structure. I leave it to the reviewer to decide whether or not that is important. (yes the tests will fail without making an exception, but is the behavior worth an exception?)

WillEngler (changes requested):
I think the singular/plural stuff is confusing and a little more involved than it needs to be. I left some suggestions on interface changes that I think could make it easier to follow and implement

garden_ai/backend_client.py
```diff
@@ -71,26 +72,39 @@ def publish_garden_metadata(self, garden: Garden):
         self._post("/garden-search-record", payload)
 
     def _get_presigned_url(
```
WillEngler, line 74: nit: this takes multiple model names so the helper method should be plural

garden_ai/client.py
```diff
@@ -458,15 +458,37 @@ def get_published_garden(self, doi: str) -> Garden:
...
         for pipeline in garden.pipelines:
-            pipeline_url_json = self.generate_presigned_urls_for_pipeline(pipeline)
-            pipeline._env_vars = {GardenConstants.URL_ENV_VAR_NAME: pipeline_url_json}
+            model_name_to_url = {}
+            for model_name in pipeline.model_full_names:
+                model_name_to_url[model_name] = next(
+                    presigned_url.url
+                    for presigned_url in all_presigned_urls
+                    if presigned_url.model_name == model_name
+                )
```
WillEngler, line 476: Not a blocker, but I feel like there's a more straightforward way to express this.

WillEngler (commented):
Awesome, this is almost there

WillEngler (approved):
LGTM!

Issue #160, opened 2023-06-06
"Could not parse {file} as a Garden pipeline. Could not execute the Python code" error masks what the real error was


WillEngler:
I put an invalid type for a step and got the error message in the title when trying to register the pipeline. We need to display more information from the underlying exception when this error is triggered.

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am trying to register a pipeline that Garden can't parse or execute, when Garden CLI throws an error at registration time, I see a helpful error message that tells me what is wrong with my code.

PR #211, opened 2023-07-11
Pipeline registration error messages

adaem2:
https://github.com/Garden-AI/garden/issues/160 

## Overview
When a user has some error in pipeline registration, there is more clear feedback on that error.

## Discussion
Provides more clear output on what error occurs when trying to register pipeline-- such as missing imports, mistyped requirements.txt file, missing pipeline object, misspelled or missing model. There are some errors that don't get caught by pipeline registration, such as non-syntax errors encapsulated in a function, which get caught later in the garden process when the steps are executed.

## Testing
I tested it by testing what "broke" the pipeline registration. 

## Documentation

Does this PR update any docs pages? Should it?
--

----
:books: Documentation preview :books:: https://garden-ai--211.org.readthedocs.build/en/211/


OwenPriceSkelly (changes requested):

garden_ai/utils/_meta.py
```diff
@@ -135,13 +138,20 @@ class _USER_PIPELINE_MODULE:
...
     exec_getsource(code_str, __main__.__dict__, local_namespace)
-    cls = local_namespace["_USER_PIPELINE_MODULE"]
 
+    cls = local_namespace["_USER_PIPELINE_MODULE"]
     # Now, one of those class attributes is going to be a Pipeline instance
     for name, value in vars(cls).items():
         if isinstance(value, Pipeline):
-            # use whatever identifier the user assigned the pipeline to in their code
+            if (
+                str(value.model_full_names)[2:-2]
```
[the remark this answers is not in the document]
OwenPriceSkelly: just remembered the point I was forgetting on our call -- another case that we should handle in the `mlmodel.Model` function is the case where the model name isn't the default model name, it's just misspelled. I think that checking to make sure there's an entry in the user's local data for the given model name is probably enough for those cases

garden_ai/utils/_meta.py
```diff
@@ -135,15 +136,21 @@ class _USER_PIPELINE_MODULE:
...
     exec_getsource(code_str, __main__.__dict__, local_namespace)
-    cls = local_namespace["_USER_PIPELINE_MODULE"]
 
+    cls = local_namespace["_USER_PIPELINE_MODULE"]
     # Now, one of those class attributes is going to be a Pipeline instance
     for name, value in vars(cls).items():
         if isinstance(value, Pipeline):
-            # use whatever identifier the user assigned the pipeline to in their code
+            if not Model.has_been_called:
+                raise PipelineLoadException(
```
OwenPriceSkelly, line 147: I think we should change this to just logging a warning instead of raising an exception, since there might be edge cases where users don't use a model in their pipeline at all (like a garden to compare a non-ML approach to an ML approach to the same scientific problem)

tests/test_filesystem_utils.py
OwenPriceSkelly, line 1: tests look great!

OwenPriceSkelly (changes requested):
Looking really good! Just one tweak and I think this is all set 🌱

garden_ai/utils/_meta.py
adaem2: Fixed, thanks!

OwenPriceSkelly (approved):
lgtm!

Issue #215, opened 2023-07-14
Epic: add "skinny pipeline" support to Garden


WillEngler:
In anticipation of Foundry wrapping Garden, we'd like to make a streamlined way for Foundry to create pipelines that just invoke models. Design doc here: https://docs.google.com/document/d/1JBCHYjWzgKXd5LAJO2ekKed0gKgaGczYIsl6njnKfN4/edit

# Assumptions:
1. Whoever picks this up should break up the work into smaller tickets first
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Closed by #227

PR #227, opened 2023-07-19
first draft support functions

### LINKED ISSUE #226: add support functions for simplified garden SDK usage
As a part of #215 the SDK must implement some convenience functions that Foundry can employ to supplant DLHub as its provider of model sharing and executing. These functions are also generally applicable as a simplified workflow for using garden.

# Assumptions:
1. There is a ticket in the Foundry repo that employs these tools and, when ready, guts DLHub.

# Acceptance Criteria
Given..., when..., then...

isaac-darling:
Part of #215 and #226

## Overview

Resolve #226 by implementing the necessary convenience functions.

## Discussion

The full context for this work is outlined in #215.

## Testing

I ran the new code in succession to my seedling end-to-end run, and it produced identical results.

----
:books: Documentation preview :books:: https://garden-ai--227.org.readthedocs.build/en/227/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -370,6 +370,8 @@ class GardenClient:
         model_name: str,
         local_path: str,
         flavor: str,
+        input_type: type,
+        output_type: type,
         requirements_file: str = "requirements.txt",
     ) -> Step:
         with open(requirements_file, "r") as f:
@@ -387,29 +389,28 @@ class GardenClient:
 
         @step
         def run_inference(
-            input_arg: object, model=Model(registered_model.full_name)
-        ) -> object:
+            input_arg: input_type, model=Model(registered_model.full_name)
+        ) -> output_type:
             return model.predict(input_arg)
 
         return run_inference
 
-    def register_minimal_pipeline(
+    def publish_minimal_pipeline(
         self,
         _step: Step,
+        garden_doi: str,
         *,
         authors: List[str],
         tags: List[str],
         title: str = None,
         description: str = None,
         requirements_file: str = "requirements.txt",
-    ) -> Pipeline:
+    ) -> Garden:
         full_name = _step.model_full_names[0]
         pipeline = self.create_pipeline(
             authors,
             title or f"Inference on {full_name}",
-            short_name=full_name[
-                full_name.find("/") + 1 :
-            ],  # just the model short_name?
+            short_name=full_name[full_name.index("/") + 1 :],
             steps=(_step,),
             requirements_file=requirements_file,
             description=description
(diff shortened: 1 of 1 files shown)


garden_ai/client.py
```diff
@@ -362,6 +364,126 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+        tags: List[str],
+        title: str = None,
+        description: str = None,
+        requirements_file: str = "requirements.txt",
+    ) -> Garden:
+        full_name = _step.model_full_names[0]
+        pipeline = self.create_pipeline(
+            authors,
+            title or f"Inference on {full_name}",
+            short_name=full_name[full_name.index("/") + 1 :],
```
WillEngler, line 413: nitpick: the user is providing us the short name as part of the simple publishing flow, but here we have to take apart a string to figure it out again

garden_ai/client.py
```diff
@@ -362,6 +364,126 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+        self,
+        *,
+        model_name: str,
+        local_path: str,
+        flavor: str,
+        input_type: type,
+        output_type: type,
+        requirements_file: str = "requirements.txt",
+    ) -> Step:
+        with open(requirements_file, "r") as f:
```
WillEngler, line 377: I'm guessing that for this flow we're more likely to be passing in requirements as a list of strings than as a file

garden_ai/client.py
```diff
@@ -362,6 +364,126 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+
+        return run_inference
+
+    def publish_minimal_pipeline(
+        self,
+        _step: Step,
+        garden_doi: str,
+        *,
+        authors: List[str],
+        tags: List[str],
```
WillEngler, line 404: There will potentially be even more metadata fields like this in the near future. (The stuff Allison and Chase are up to.) Should we keep growing this argument list for every field?

WillEngler (commented):
Left some quick thoughts - good job getting a prototype up quickly

garden_ai/client.py
isaac-darling: The logical response being "No," but stick with me for a second. The obvious response is to setup some **kwargs and let Python do the rest; but then you say "what if the user provides nonsense?" My solution, however, is not to make a schema, which will cost dev hours down the line as users fail to use it properly anyway and bugs inevitably arise, and rather to put our foot down and say this is a minimal example in every sense of the word--and if something arises that is so critical it becomes a minimum requirement, then the trouble to add an argument to this function is negligible.

garden_ai/client.py
```diff
@@ -362,6 +364,114 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
         local_data.put_local_pipeline(registered)
         return func_uuid
 
+    def _create_minimal_step(
```
WillEngler, line 367: Can you add dcostring comments to these new methods? And maybe even add a sort of section delimiter to show to someone scrolling through that they're related?

garden_ai/client.py
```diff
@@ -362,6 +364,114 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+        self,
+        *,
+        model_name: str,
+        local_path: str,
+        flavor: str,
+        input_type: type,
+        output_type: type,
+        requirements: List[str],
```
WillEngler, line 375: Having slept on it ... this is just too many parameters. And right below you bundle them up into this nice LocalModel object that is a logical grouping of fields that comprise an ML model. Could this method take a LocalModel? 

Please have another pass at this and think about how to use our existing Pydantic models or even come up with new ones as needed to make a more understandable interface that will take less maintenance and upkeep.

WillEngler (commented):
Strongly nudging (ie, telling) you to reconsider the really long parameter list approach. We don't need a "schema" per se. But bundling up conceptually related fields into struct-like data structures where appropriate and reusing those data structures around the code base is big for maintainability and readability.

garden_ai/client.py
OwenPriceSkelly: just to chime in on this, I think that you are basically right when you say the obvious thing to do is setup some **kwargs and let ~python~ pydantic do the rest. 

See for example the way you use the `create_pipeline` method here -- when `create_pipeline` was written, we had no concept of the `short_name` field. Cut to now, a world where you need `short_name` to satisfy the `RegisteredPipeline` schema. Notice how you were able to easily specify `short_name=model_name` at the call site but we didn't need to modify the definition of `create_pipeline` in tandem with the changing schema? 

tl;dr object/dataclass/schema/whatever is already the place where we can/ought to encode "what's the absolute minimum"

garden_ai/client.py
```diff
@@ -362,6 +364,110 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+        local_model = LocalModel(
+            model_name="dendrite_segmentation",
+            flavor="tensorflow",
+            local_path="model.h5",
+            user_email=client.get_email()
+        )
+        my_pipeline = client.add_simple_model_to_garden(local_model, "10.1234/doi-here",
+                                                        input_type=np.ndarray,
+                                                        output_type=np.ndarray,
+                                                        authors=["Monty Python", "Guido van Rossum"],
```
WillEngler, line 425: heavy hitters

garden_ai/client.py
```diff
@@ -362,6 +364,110 @@ def register_pipeline(self, pipeline: Pipeline, container_uuid: str) -> str:
...
+                                                        pip_dependencies=["tensorflow"],
+                                                        tags=["materials science", "computer vision"]
+        )
+        """
+        # caller needs to subclass and manually set required attribute of the field to False to avoid errors and maintain ignorance
+        local_model.user_email = self.get_email()
```
WillEngler, line 431: Or there could be some wrapper that grabs the email if they don't provide one

WillEngler (approved):
LGTM, thanks for going a few rounds on this

WillEngler (approved):

PR #236, opened 2023-07-25
Container caching

### LINKED ISSUE #217: naive container uuid caching
_As a garden developer/power-user, I may frequently re-register a pipeline with only minor tweaks that don't impact the requirements for the pipeline. Despite this, I still need to wait for a whole new container to build every single time I re-register._

# Assumptions:
1. we have a record in local data we can use as a mapping from requirement sets -> container uuids
    - implementation doesn't need to be fancy but obvious stuff like order shouldn't matter if the same requirements are pinned 
    - a pinned requirement and an unpinned or differently-pinned requirement should not be considered the same 
3. pipeline registration checks that mapping before issuing the expensive container build request 
4. there should be a CLI option to force a new container build 

# Acceptance Criteria
Given a pipeline with the same exact requirements as an already-registered pipeline, they are registered against the same container

isaac-darling:
`Fixes #217`

## Overview

Adds caching to the container building process. Will work for pipelines sharing a DOI and all dependencies.

## Discussion

Is this actually usable behavior? Does the content of the container not also depend on the content of the pipeline itself? If it doesn't (just needs uuid or otherwise), fantastic, but if it does, this may not be a usable feature except in very specific cases.

## Testing

I ran from scratch several builds locally, and the caching worked as expected.

## Documentation

There is additionally a change in the return type of `client.register_pipeline`, just for fun. I'll fix what breaks in relation to that.

----
:books: Documentation preview :books:: https://garden-ai--236.org.readthedocs.build/en/236/

### DIFF
--- a/garden_ai/app/pipeline.py
+++ b/garden_ai/app/pipeline.py
@@ -352,11 +352,9 @@ def register(
         console.log(f"Could not parse {pipeline_file} as a Garden pipeline. " + str(e))
         raise
     with console.status(
-        "[bold green]Building container. This operation times out after 30 minutes."
+        "[bold green]Building container and registering pipeline. This operation times out after 30 minutes."
     ):
-        container_uuid = client.build_container(user_pipeline)
-    console.print(f"Created container {container_uuid}")
-    func_uuid = client.register_pipeline(user_pipeline, container_uuid)
+        func_uuid = client.register_pipeline(user_pipeline)
     console.print(f"Created function {func_uuid}")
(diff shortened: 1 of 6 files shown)


OwenPriceSkelly:
notes from syncing up, for bookkeeping:
- change the cache signature from `doi : container_uuid, requirements` to `hash(frozenset(requirements)) : container_uuid`
- maybe print a cheerful message to the user so they know that they got to skip the build step 
- add some quick a happy path/sad path unit tests that assert that the expensive container build method was not called / was called once, respectively

OwenPriceSkelly (approved):
LGTM!

PR #244, opened 2023-07-31
Pytorch tensor input support

### LINKED ISSUE #206: Pytorch models should accept tensors
The way we have things hooked up right now, PyTorch models only accept dataframes as input. As a user, I expected them to accept `torch.Tensor`s because the underlying model I trained accepted tensors.

MLFlow has ways to allow this. It's just that our naive first draft of model wrapping doesn't take advantage of it. Here's a blog post we can work off of to get this working: https://www.databricks.com/blog/2021/03/18/tensor-input-now-supported-in-mlflow.html

# Assumptions:
1.
2.

# Acceptance Criteria
Given I register a torch model with Garden that takes a tensor, when I remotely execute it with a tensor, then it works.

MaxTuecke:
Fixes #206

## Overview

Pytorch models currently do not accept inputs of type `torch.tensor`. This is because when loading torch models, Garden uses`mlflow.pyfunc.load_model`, which makes a mlflow model that can accept inputs of type `np.ndarray` and `pd.dataframe` but not `torch.tensor`. However, if Garden loads torch models with `mlflow.pytorch.load_model`, the model accepts inputs of type `torch.tensor` but not `np.ndarray` and `pd.dataframe`.

For the time being, all torch loads have been changed to `mlflow.pytorch.load_model` and assume torch users will always be using inputs of type `torch.tensor`. 

In the future, we might want to consider adding our own model signatures so users can specify input types of their models.

----
:books: Documentation preview :books:: https://garden-ai--244.org.readthedocs.build/en/244/


garden_ai/_model.py
```diff
@@ -99,13 +99,32 @@ def clear_mlflow_staging_directory():
...
         if self._model is None:
             download_url = self.get_download_url(self.full_name)
             local_model_path = self.download_and_stage(download_url, self.full_name)
-            self._model = mlflow.pyfunc.load_model(
-                local_model_path, suppress_warnings=True
-            )
+            mlflow_yaml_path = local_model_path + "/MLmodel"
+            with open(mlflow_yaml_path, "r") as stream:
+                mlflow_metadata = yaml.safe_load(stream)
+            mlflow_load_strategy = mlflow_metadata["metadata"]["garden_load_strategy"]
```
WillEngler, line 110: Will this give a KeyError for old models? (Not a deal breaker, we can do another DB wipe. We just have to be sure to do it.)

WillEngler (approved):
LGTM

PR #246, opened 2023-08-04
Sklearn preprocessor support 

### LINKED ISSUE #218: Support scikit-learn preprocessors
The mlflow pyfunc flavor works for a lot of cases where it can wrap a model with a `.predict()` call. One place where it can't is scikit-learn preprocessors like StandardScaler. In the _Model class, we always load models with `mlflow.pyfunc.load_model`. This does not work for sklearn preprocessors. But loading them with `mlflow.sklearn.load_model` does seem to work. 

It would be good for garden to support this. At minimum, MAST-ML makes heavy use of sklearn preprocessors.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

MaxTuecke:
Fixes #218 

## Overview
Allows sklearn pipelines and preprocessors to be upload as models to Garden. 

* Redefines `__getattr__` in `_Model` to proxy all attributes of the stored `_model` to the `_Model` class. This allows users to be able to call the internal `_model` methods using their `_Model` object.
* Changes mlflow load for sklearn models from `mlflow.pyfunc.load_model` to `mlflow.sklearn.load_model`.
* Adds new metadata `garden_load_strategy` to mlflow upload.
* Adds unit tests for loading sklearn models.

----
:books: Documentation preview :books:: https://garden-ai--246.org.readthedocs.build/en/246/

### DIFF
--- a/garden_ai/_model.py
+++ b/garden_ai/_model.py
@@ -121,7 +121,6 @@ class _Model:
                 )
             elif mlflow_load_strategy == "sklearn":
                 self._model = mlflow.sklearn.load_model(local_model_path)
-                # self.do_passthough = True
             elif mlflow_load_strategy == "pytorch":
                 # Load torch models with mlflow.torch so they can taketorch.tensors inputs
                 # Will also cause torch models to fail with np.ndarrays or pd.dataframes inputs
@@ -130,7 +129,6 @@ class _Model:
                 self._model = self._TorchWrapper(
                     mlflow.pytorch.load_model(local_model_path)
                 )
-                # self.do_passthough = True
             else:
                 raise Exception(
                     f"Invlaid garden_load_strategy given: {mlflow_load_strategy}"
@@ -164,7 +162,7 @@ class _Model:
         """
         return self.model.predict(data)
 
-    # Dill serialization and breaks with __getattr__ if missing this.
+    # Dill serialization breaks using new __getattr__ if missing this.
     def __getstate__(self):
         return self.__dict__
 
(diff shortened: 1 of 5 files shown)


garden_ai/_model.py
```diff
@@ -105,13 +105,23 @@ def _lazy_load_model(self):
...
-
-            if mlflow_load_strategy == "mlflow.pytorch":
+            try:
+                with open(mlflow_yaml_path, "r") as stream:
+                    mlflow_metadata = yaml.safe_load(stream)
+                mlflow_load_strategy = mlflow_metadata["metadata"][
+                    "garden_load_strategy"
+                ]
+            except Exception:
+                # Default to mlflow.pyfunc if can't find garden_load_strategy
```
WillEngler, line 115: Appreciate this

garden_ai/_model.py
```diff
@@ -152,10 +162,31 @@ def predict(self, data):
...
+            if attr in self.__dict__:
+                # Use self definition of attr
+                return self.__getattribute__(attr)
+            else:
+                # self does not have definition of attr
+                # Search _model instead
+                return self._model.__getattribute__(attr)
+        else:
+            # _model is None, lazy load and try again.
+            self._lazy_load_model()
```
WillEngler, line 183: Are there cases where we could hit this branch before we'd normally need to load the model? Probably not in practice? If so this would be fine. But if there are then I'd make this a little more restrictive so we try 

```
if attr in self.__dict__:
                # Use self definition of attr
                return self.__getattribute__(attr)
```
before loading the model

WillEngler (approved):
Overall looks good to me. Because there are some Dill serialization workarounds I'd have @OwenPriceSkelly take a quick look before merging for situational awareness on his end.

OwenPriceSkelly (approved):
LGTM 🌱 just for bookkeeping, held a sync review and talked through changes, minor nits around catching bare `Exception`, otherwise concluded that anything hacky-feeling here is reasonable and morally justified. nice work!

PR #247, opened 2023-08-07
add published garden

### LINKED ISSUE #127: new `RegisteredGarden` entity vs plain `Garden`
_Analogous to #113, we're starting to want/need distinct objects for the distinct "sources of truth" that might be used to instantiate a garden, namely `Garden`s that came from local data vs `RegisteredGardens` that came from a Globus Search result._

In other words: `RegisteredGardens` are the border between the sdk and Globus Search. Regular `Garden` instances correspond to a record in local storage, `RegisteredGarden` instances correspond to a record in Globus Search. 

# Assumptions:
1. You can't publish a `Garden`, only `RegisteredGarden` (i.e. an instance of the latter should be what gets serialized, even if it's just `RegisteredGarden(**garden.expanded_metadata())` before publishing)
2. Successful search results become `RegisteredGarden`s, not `Garden`s.
    - helper functions `globus_search.garden_search.get_remote_garden_by_{uuid, doi}` should return `RegisteredGarden` objects
3. `RegisteredGarden` objects are for "deep" metadata; their json/dict representations contain the full metadata of their `RegisteredPipeline`s, where `Garden`s have just the `pipeline_ids` with helper methods like `expanded_metadata()` or `_collect_pipelines()`. 
4. `Garden` objects should only be able to instantiate pipelines by fetching their metadata according to the `pipeline_ids` from local_storage 

# Acceptance Criteria
Given: a globus search record of a garden, 
When: that record was published, it was published from a `registered_garden.json()`
When: that record is downloaded, then it instantiated a `RegisteredGarden` 

## other notes/discussion
- `RegisteredGarden` could also override `__dir__` for tab-completion in notebook/repls
- Might want to replicate the pattern of the `RegisteredPipeline.from_pipeline` classmethod to construct a `RegisteredGarden` from a plain Garden.  
- upshot of 3 is that `RegisteredGardens` are free to have a `pipelines: List[RegisteredPipeline]` attribute instead of `Garden`s read-only `@property` approach 
- This story ended up being pretty opinionated w/r/t implementation, I think it's probably still valuable but no need to take it as gospel

isaac-darling:
`Fixes #127 `

## Overview

Add a `PublishedGarden` object to clarify the roles a garden can play as both a work in progress and a record.

## Testing

In addition to the tests here, the full ETE tests will be run.

## Documentation

This adds the new object to the docs through the md yaml, possible fixes required.

----
:books: Documentation preview :books:: https://garden-ai--247.org.readthedocs.build/en/247/


garden_ai/gardens.py
```diff
@@ -260,6 +218,185 @@ def expanded_json(self) -> JSON:
...
+    tags: List[str] = Field(default_factory=list, unique_items=True)
+    version: str = "0.0.1"
+    pipelines: List[RegisteredPipeline] = Field(...)
+    pipeline_ids: List[str] = Field(...)
+    pipeline_aliases: Dict[str, str] = Field(default_factory=dict)
+    pipeline_names: List[str] = Field(...)
+    _env_vars: Dict[str, str] = PrivateAttr(default_factory=dict)
+
+    @root_validator(pre=True)
+    def verify_pipeline_integrity(cls, values):
```
OwenPriceSkelly, line 348: this smells odd -- why do `PublishedGardens` need `pipeline_ids` and `pipeline_names` as attributes when they already have full copies of their `RegisteredPipelines`? I think we shouldn't duplicate that state at the garden level when it's already available at the pipeline level. that way we don't need to handle them falling out of sync like this validator is anticipating

garden_ai/gardens.py
```diff
@@ -298,61 +434,13 @@ def datacite_json(self) -> JSON:
             else None,
         ).json()
 
-    def validate(self):
```
OwenPriceSkelly, line 301: love this diff 🧑‍🍳🤌

garden_ai/gardens.py
```diff
@@ -367,29 +455,6 @@ def __dir__(self):
         # this gets us jupyter/ipython/repl tab-completion of pipeline names
         return list(super().__dir__()) + self.pipeline_names
```
OwenPriceSkelly, line 456: since I'm thinking we should cut the `self.pipeline_names` attribute:
```suggestion
        pipeline_names = [self.pipeline_aliases.get(p.doi) or p.short_name for p in self.pipelines]
        return list(super().__dir__()) + pipeline_names
```

tests/test_local_db.py
```diff
@@ -56,6 +56,22 @@ def test_local_storage_model(mocker, database_with_model, second_draft_of_model)
...
+
+    # mock fetch and new garden creation operations
+    mocker.patch(
+        "garden_ai.client.GardenClient.get_published_garden",
+        return_value=PublishedGarden.from_garden(garden_all_fields),
+    )
+    mocker.patch("garden_ai.client.GardenClient.create_garden", return_value=None)
+
+    garden_client.clone_garden(garden_all_fields.doi, silent=True)
+
```
OwenPriceSkelly, line 71: should add a quick comment here explaining the behavior we're asserting, i.e. cloning a published garden should rehydrate local storage with the pipeline metadata copied from the garden

OwenPriceSkelly (changes requested):
this is looking really good, extremely excited to see this merged. just a handful of nits but I think they'll be easily picked 🐜🌱

OwenPriceSkelly (approved):
LGTM! really nice work on this, merge away 🫡

Issue #37, opened 2023-01-10
Hello, Model Registry


BenGalewsky:
_As a garden user I want a registry to publish my models to and read from_

# Assumptions:
1. Will be a deployment of MLFlow tracking server on ECS behind an ALB using OAuth2 proxy. See [this article](https://getindata.com/blog/deploying-secure-mlfow-aws/) for details.
2. The MLFlow server will [proxy access to S3 storage](https://mlflow.org/docs/latest/tracking.html#scenario-5-mlflow-tracking-server-enabled-with-proxied-artifact-storage-access)

# Acceptance Criteria
Given I visit the URL for the deployed tracking server When I follow the GlobusAuth flow Then I can see the tracking server

Given I visit the URL for the deployed tracking server When I provide invalid Globus credentials Then I am not allowed to see the tracking server

WillEngler:
Implemented here https://github.com/Garden-AI/garden-backend/pull/5

Issue #216, opened 2023-07-17
Allow `pytorch` users to specify python files to include with model registration


OwenPriceSkelly:
_As a `pytorch` user, a common pattern is to implement and train my model as a custom `torch.nn.Module` subclass. We will need that class definition to be present in order to deserialize the model on the other end._

For example, the `BraggNN` class is defined here: https://github.com/blaiszik/dlhub_publish/blob/main/Bragg_model_code_notebook/model.py and that class definition is needed to register the `.pth` file in the same directory. 

# Assumptions:
1. `mlflow.pytorch.log_model` accepts a `code_paths` kwarg that can contain a `"path/to/my/class/definition.py"`, which will be included as part of the model at registration time
    - i.e. mlflow can handle this fine, we just need to include the file defining the user's custom class with the `model.pth` when we upload/register the model
2. `LocalModel` class has a field for code_paths that can be used to register; `RegisteredModel` won't need this field 

# Acceptance Criteria
- if I'm trying to register a pytorch model, I can specify additional python files that should be uploaded as part of the model, whether from the command line or the client 
- this behavior is documented and tested

PR #248, opened 2023-08-07
Pytorch model registration

adaem2:
Link to any related issues, companion PRs, or blockers. eg, `Fixes #37`
https://github.com/Garden-AI/garden/issues/216
## Overview

When a Pytorch model has extra python file dependencies, a user can link these files for mlflow to process.
e.g. garden-ai model register model_name model_path pytorch -e python_file_path

## Discussion

Any additional context that you think is important for reviewers or future users/contributors looking at this PR.

## Testing

How did you test this? If a reviewer pulled this branch down and wanted to kick the tires, how would they do so?
Tested it on the BraggNN model example. 
## Documentation

Does this PR update any docs pages? Should it?

----
:books: Documentation preview :books:: https://garden-ai--248.org.readthedocs.build/en/248/


OwenPriceSkelly (commented):
This looks great! ~only question is about the new pytorchtest.pth and torch.py fixtures -- are those used anywhere? the unhappy path unit test looks great, any reason not to have a happy path test that does the same with those fixtures?~ edit: just saw your comment in slack; go ahead and include the test again but mark it as an "integration" test. That way it doesn't fire as part of the regular GH actions but we still have it around and it'd run if we just run pytest by itself locally

OwenPriceSkelly (approved):
beautiful. LGTM!

PR #249, opened 2023-08-08
add all container spec vars to cache tag

isaac-darling:
## Overview

Adds every parameter to the container build operation as a component of the cache tag.

## Testing

I tested the cache locally and it worked.

----
:books: Documentation preview :books:: https://garden-ai--249.org.readthedocs.build/en/249/


OwenPriceSkelly (approved):
LGTM!

PR #254, opened 2023-08-14
implement container shell command

### LINKED ISSUE #170: Extend debug shell with optional Docker build support
The debug shell is great for letting users test their pipelines while making no assumptions about additional tools to install. But it will probably be useful to have a debugging experience that is closer to our Dockerized production environment.

Extend garden-ai pipeline shell (or make a new command) to build a Docker image based on the ContainerSpec that would be passed to Container Service and run the pipeline within that container.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

### LINKED ISSUE #171: Debugging helper for containers built by Container Service
Containers built by Container Service are publicly accessible. We might want to pull them down to poke around and debug. But it will be unintuitive to use off the shelf. (eg, the entrypoint is not to an interactive shell.) Do whatever combination of documenting or adding new shell utilities are needed to make it straightforward to debug containers built by container service on a dev's machine.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

isaac-darling:
Fixes #171 

## Overview

Rather than creating a venv for the user, spin up the actual container from the container service.

## Testing

I ran the command with and without the cleanup option and everything worked as expected.

----
:books: Documentation preview :books:: https://garden-ai--254.org.readthedocs.build/en/254/


garden_ai/app/pipeline.py
```diff
@@ -363,101 +366,64 @@ def register(
...
 
-        # Start Python shell in the virtual environment with the pipeline file
-        print("Starting Garden test shell. Loading your pipeline one moment...")
-        python_command = os.path.join(temp_dir, "bin", "python")
-        subprocess.run([python_command, "-i", pipeline_file])
+        container_info = client.compute_client.get_container(
+            container_uuid or cached_uuid, "docker"
+        )
+        location = container_info["location"]
+        location = location[location.index("bengal1") :]  # remove docker.io domain
```
OwenPriceSkelly, line 416: do we need to remove the docker.io domain? I thought that `docker pull 'docker.io/bengal1/bunchanumbers:latest'` would be the same as `docker pull 'bengal1/bunchanumbers:latest'` by default

OwenPriceSkelly: if we do need/want to strip it, it's better to hardcode the docker.io part we want to strip than ben g's username, since it's more likely that we change the user than the registry itself

OwenPriceSkelly (changes requested):
This looks very tidy, nice and easy to follow. High-level points I'm happy to go back/forth on if you feel strongly: 
- I think we should keep the old `shell` command around for people who don't have docker, since that doesn't require anything but python. it's also possible that the pipeline is failing to even register a container due to bad/incompatible requirements, and `pipeline shell` is worth having around as a smoke test for that 
- I think this command should be `pipeline container` or something to indicate that it's more involved (and more robust, hypothetically) than just what e.g. the analogous `poetry shell` would do 
- It would probably be especially convenient to accept either a `pipeline.py` file as well as the doi/uuid, since the use case we're imagining is one where a pipeline needs to be debugged and the feedback loop of: 
    edit `pipeline.py` -> `pipeline container pipeline.py` -> edit `pipeline.py` 
is a lot tighter than: 
    edit `pipeline.py` -> `pipeline register pipeline.py` -> `pipeline container [doi]` -> edit `pipeline.py`
and the container is likely not changing in between. If the user does change the requirements s.t. there's no cache hit I think we just ask them to re-register

OwenPriceSkelly (changes requested):

garden_ai/app/pipeline.py
```diff
@@ -461,6 +465,101 @@ def shell(
...
+                with console.status(
+                    "[bold green]Building container. This operation times out after 30 minutes."
+                ):
+                    cached_uuid = client.build_container(user_pipeline)
+
+        container_info = client.compute_client.get_container(
+            container_uuid or cached_uuid, "docker"
+        )
+        location = container_info["location"]
+
```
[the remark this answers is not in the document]
isaac-darling: docker has its own progress bars, which will still show

OwenPriceSkelly (approved):
lgtm!

PR #258, opened 2023-08-22
Add field to Pipeline that lets user specify container uuid

### LINKED ISSUE #255: Add ability to specify container ID in CLI pipeline registration
To register models from Eliu's framework, I needed to build a custom container without using Container Service. But there was no way to specify the container uuid from the CLI. So I hardcoded the id in to my local copy of Garden. It would be great to make this a real feature that people who are not devs on Garden could use.

# Assumptions:
1. Container Service will still be the default, but there will be a way to specify a Globus Compute container ID from the command line 
2.

# Acceptance Criteria
Given I am registering a pipeline that needs a custom container, when I provide the container id on the command line, then the pipeline is registered to run on that container, as opposed to attempting a container service build.

WillEngler:
Closes #255

## Overview

Adds a new optional field to the Pipeline object where a user can specify the UUID of a registered container in Globus Compute.

## Discussion

I'm not advertising this far and wide in the docs or any console messages just yet. Still thinking through how we really want to handle cases where we need custom containers, but this is a good escape hatch for now.

## Testing

I added a unit test to confirm that the UUID gets passed into the function registration process as I expect. I also did some manual testing where I registered a pipeline using this parameter and confirmed that it was registered with the specified container.

## Documentation

I'm keeping it sneaky for now.

----
:books: Documentation preview :books:: https://garden-ai--258.org.readthedocs.build/en/258/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -430,7 +430,6 @@ class GardenClient:
             else:
                 container_uuid = self.build_container(pipeline)
 
-        print(container_uuid)
         func_uuid = register_pipeline(self.compute_client, pipeline, container_uuid)
         pipeline.func_uuid = UUID(func_uuid)
         registered = RegisteredPipeline.from_pipeline(pipeline)
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -413,7 +413,9 @@ class GardenClient:
         return built_container_uuid
 
     def register_pipeline(self, pipeline: Pipeline) -> RegisteredPipeline:
-        container_uuid = pipeline.container_uuid
+        container_uuid = (
+            str(pipeline.container_uuid) if pipeline.container_uuid else None
+        )
         if container_uuid is None:
             cache = _read_local_cache().get(
                 get_cache_tag(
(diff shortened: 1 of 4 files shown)


tests/test_client.py
```diff
@@ -196,3 +197,26 @@ def test_client_credentials_grant(cc_grant_tuple):
...
+def test_register_pipeline_with_specified_uuid(
+    mocker, garden_client, pipeline_toy_example
+):
+    CONTAINER_UUID = "3dc3170e-2cdc-4379-885d-435a0d85b581"
+    pipeline_toy_example.container_uuid = CONTAINER_UUID
+
+    # Mock register_pipeline so we can see if the container uuid gets passed in
+    mocker.patch(
+        "garden_ai.client.register_pipeline",
+        return_value="a4742a1b-966a-4913-a369-fb6555b8bfb9",
```
OwenPriceSkelly, line 211: not that we use it, but wouldn't the return value here now be a `RegisteredPipeline`?

WillEngler: Confusingly, this is mlmodel.register_pipeline being mocked. Just mocking it at the import site. mlmodel.register_pipeline does indeed return a uuid

garden_ai/client.py
```diff
@@ -429,6 +430,7 @@ def register_pipeline(
             else:
                 container_uuid = self.build_container(pipeline)
 
+        print(container_uuid)
```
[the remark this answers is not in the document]
WillEngler: d'oh

WillEngler:
Sweet, removed the print statement and I'll merge after tests pass.

PR #259, opened 2023-08-23
Add client method to upload models from memory without serializing to disk first

### LINKED ISSUE #232: New client method to register model that is already in memory
Right now, the `register_model` Garden client method and the `garden-ai model register` command that wraps it assumes that you have a serialized model on disc. Working with [Eliu's notebook](https://colab.research.google.com/drive/1viGlpUIGmRW-kzqERj4fGzX9udEXVrXw?usp=sharing#scrollTo=dkLCvw1xM3vo), I found myself wanting to use his existing code to load his models into memory and then invoke the Garden client on those. But as it stands I have to figure out how to serialize them first.

This feature could also be useful for publishers to incorporate Garden into training workflows. (Adding some step at the end of their workflow to push the model they've just trained to Garden without needing to write it to disc first.) 

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Closes #232 

## Overview

Adds client method that is meant to be used from a notebook context or similar to upload a model that's in memory without needing to serialize and deserialize it.

## Discussion

As a byproduct this breaks up the monster `stage_model_for_upload` function in mlmodel.py. The original behavior should be exactly preserved though, bar a name change.

## Testing

- Updated existing tests
- Added an integration test that invokes the new client method
- Imported this branch in the context of Hyun/Eliu's notebook to test the new client method "in the wild". Successfully uploaded `willengler@uchicago.edu)torchmdnet-ethanol` model.

## Documentation

Added example to tutorial showing how to use from a notebook or similar context.

----
:books: Documentation preview :books:: https://garden-ai--259.org.readthedocs.build/en/259/

### DIFF
--- a/docs/Models.md
+++ b/docs/Models.md
@@ -7,4 +7,4 @@
 			- LocalModel
 			- ModelMetadata
 			- DatasetConnection
-			- stage_model_for_upload
+			- stage_model_from_disk
--- a/docs/user_guide/tutorial.md
+++ b/docs/user_guide/tutorial.md
@@ -123,6 +123,8 @@ garden_client.register_model_from_memory(my_model, model_meta)
 
 ```
 
+In this example the name of the registered model to invoke will be `willengler@uchicago.edu/torchmdnet-ethanol`.
+
 ### Register a Pipeline
 
 Once we're satisfied with our pipeline's definition in `pipeline.py`, we can register the pipeline, allowing us to:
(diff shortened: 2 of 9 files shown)


OwenPriceSkelly (approved):
LGTM! but can we add a line to the new docs section/example that mentions that the full model name to use would be`"willengler@uchicago.edu/torchmdnet-ethanol"`?

WillEngler:
> but can we add a line to the new docs section/example that mentions that the full model name to use would be"willengler@uchicago.edu/torchmdnet-ethanol"?

Gladly! I'll get that in and then merge.

PR #277, opened 2023-09-15
Publish prototype

### LINKED ISSUE #274: Publication prototype: Run Globus Compute function in the context of a baked notebook container.
To build on the work in #272, prove out that we can execute a Globus Compute function in the context of a "baked notebook" container.

This will take a few steps, several of which can be fudged. 
1. We will need to register one or more containers built with `garden notebook bake` with Globus Compute. For prototyping purposes, I think it's OK if this is done as a one-off. We don't need to automate this in code. The real flow would involve building an Apptainer container remotely on a Globus Compute endpoint. Because this is just proving out some ideas with the old Docker/Kubernetes flow, no need to lock down something we're very likely never going to use.
2. Register a Globus Compute function from the container notebook locally. This is part of the flow we would incorporate. To prove out the concept, just registering a function from the `notebook debug` shell would suffice. But it would be pretty sick if we implemented a `@garden.pipeline()` decorator or similar and a new notebook subcommand could register the function in the context of the notebook container.
3. Run it! It doesn't have to be in a Garden for this proof of concept. Just some Globus Compute function that's invokable on our DLHub GC endpoint that successfully runs the function in the baked notebook context within the container.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

isaac-darling:
Fixes #274 

## Overview

Adds globus compute to the picture, for remote execution.

## Discussion

This is mostly here as documentation of the effort to resolve issue #274, as a fully fleshed out implementation relies on several points of contact and decisions that have yet to be made.

## Testing

Tested manual configuration of the functionality successfully.

----
:books: Documentation preview :books:: https://garden-ai--277.org.readthedocs.build/en/277/


garden_ai/app/notebook.py
```diff
@@ -96,6 +98,59 @@ def bake(
...
+    import subprocess
+
+    # overwrite any init process that utilized an unsupported dill version
+    subprocess.run(["python", "-m", "pip", "install", "dill==0.3.5.1"])
+
+    import dill
+
+    dill.load_session("session.pkl")
+
+    # don't ask how I figured it out, I won't say.
```
WillEngler, line 111: 🤐

WillEngler (approved):
LGTM!

WillEngler:
Not sure what's up with the test failures though 🤔

WillEngler (approved):
LGTM!

PR #288, opened 2023-09-29
ETE Test Updates

MaxTuecke:
## ETE Changes
- Fixed container caching
- Added `constants.py`
- Added sklearn preprocessor model to testable flavors
- Added pytorch specific pipeline templates to do `model.eval()` and `torch.no_grad()` before predicting
- Added support for testing subset of model flavors
- Added a check that the results equal the expected results for all models at the end of the remote execution 
- Added new 'fresh' option for globus-compute-endpoint argument that creates a new endpoint, tests remote execution with that endpoint, and deletes after test is done. Test times out significantly less frequently when using this new endpoint instead of the DLHub endpoint.
- Added new arguments to allow users to test custom models/pipelines 
- Added new git secrets arguments for local runs
- Changed the full test to no longer test tensorflow, at least for now (tf containers are currently broken )
- Changed the full test to use cached containers rather than build new ones (building new containers is broken for now)
- Changed full test to use fresh endpoints rather than DLHub one
- Changed skinny test to use sklearn-preprocessor
- Specified exact Python version for both full and skinny tests
- Changed tensorflow and torch models out for faster ones
- Updated --help and 'QUICK USE' to better explain how to use the test

## Other Changes
- Added `load_strategy` for torch models in `mlmodel.py`
- Fixed `__getattr__` in `_model.py` to proxy attribute lookups correctly for `_TorchWrapper` objects

----
:books: Documentation preview :books:: https://garden-ai--288.org.readthedocs.build/en/288/


WillEngler (approved):
Great additions! I did a little tweaking to pass CI and we're good to merge now.

PR #289, opened 2023-10-02
allow the prototype to run on apple silicon

isaac-darling:
## Overview

There is a missing library in the conda build step of the docker image when run on arm. The fix is to simply tell docker to emulate x86 regardless.

## Testing

I do not have an arm processor, so testing is needed on a machine that does.

----
:books: Documentation preview :books:: https://garden-ai--289.org.readthedocs.build/en/289/


WillEngler (approved):
LGTM. I haven't tested this personally, but looser standards on the prototype stuff. I'll probably run this myself tomorrow and make tweaks if I see issues. (But from my own experience building containers on my apple silicon Mac, this is indeed the right argument to pass so I'm decently confident you got it right.)

isaac-darling:
I did some debugging with Jennifer to modest success (didn't run through Garden for her, but using her image I did successfully). So at least this part of it is tested, albeit with lingering questions.

PR #290, opened 2023-10-04
Auth flow link fixes

### LINKED ISSUE #287: Can we make the Globus Auth auth flow link clickable in notebooks?
When you go through GardenClient init in a notebook, you can't click on the generated link for the authorization flow. You have to paste it in to a browser bar. This is annoying for all browsers, but for Safari in particular it seems to do something that mangles the URL, adding in a space character.

Can we output the link in a way that is clickable in Colab and Jupyter? Making it easy for everyone and avoiding the bug for Safari users. If not, maybe we can get some Globus SDK help.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

MaxTuecke:
Fixes #287

## Overview

Makes the Globus auth flow links clickable in notebooks and Google Colabs.

## Description

During the Globus auth flow, adds a check to see if Garden is running in a notebook and if it is, displays url as HTML with `IPython.display` instead of printing it. Also fixes Safari mangling the url as it is not printing the url anymore.

----
:books: Documentation preview :books:: https://garden-ai--290.org.readthedocs.build/en/290/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -204,7 +204,7 @@ class GardenClient:
             self._display_notebook_link(
                 authorize_url
             )  # Must display url as html to render properly in notebooks
-        except NameError as e:
+        except NameError:
             print(
                 f"Authenticating with Globus in your default web browser: \n\n{authorize_url}"
             )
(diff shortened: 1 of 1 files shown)


WillEngler:
Nice! This is one of those fixes that has huge ROI. Not that big of a code change but big UX win.

Issue #284, opened 2023-09-25
Make garden-ai point to production by default, including production DataCite account


OwenPriceSkelly:
_We need to make sure that our dev/prod environments use the corresponding datacite test/production endpoints to generate or update dois_
Related to #154 and #119 
# Assumptions:
1. We have the prefix, account ID, and password for production DataCite in aws secrets
2. Backend terraform has been configured to grant secrets access to dev/prod requests appropriately 

# Acceptance Criteria
Given: Some metadata and a request for a DOI
When: hitting the garden dev api
Then: datacite test endpoint generates the doi

When: hitting the garden prod api
Then: real datacite generates the doi

OwenPriceSkelly:
update: all that's "blocking" this is putting the real datacite creds in aws secrets (the backend code is already in place), we've just decided we're not actually at the point of needing/wanting real DOIs quite yet

ascourtas:
TODO: add this to a README somewhere so it's documented and we can closet this issue

Issue #283, opened 2023-09-25
wipeable data stores -- MLFlow 


OwenPriceSkelly:
Part of #154 
# Assumptions:
1. We have separate s3 buckets for storing dev and prod models

# Acceptance Criteria
Given: I've set GARDEN_ENV=dev
When: I stage and upload a model
Then: I'm targeting the dev s3 bucket

Given: I've set GARDEN_ENV=dev
When: I'm remotely executing a pipeline
Then: the presigned URLs target the dev bucket

likewise for leaving GARDEN_ENV unset or GARDEN_ENV=prod

Issue #282, opened 2023-09-25
wipeable data stores -- Search


OwenPriceSkelly:
_We need to be able to toggle using a separate globus search index for dev/prod gardens_
Part of #154 
# Assumptions:
1. There exists a second globus search index to point to for dev purposes 

# Acceptance Criteria
Given: I've set GARDEN_ENV=dev
When: I publish a garden
Then: only the dev index is written

Given: I've left GARDEN_ENV unset
When: I search for a garden
Then: only results from the prod index are returned (i.e. no test gardens appear)

PR #297, opened 2023-10-09
154 prod dev environment toggle

### LINKED ISSUE #154: Add ability to point Garden CLI to "wipeable" data stores
For both ad hoc developer testing and automated testing, we would like to be able to point to dev versions of our data stores. That way we don't pollute the production views that our end users will see. We don't want search results to be mostly spam from dev testing.

We would also like the ability to wipe out all the state in the dev data stores when it's either too cluttered or we need to make some breaking change to how we are persisting data.

There are 3 places we persist data outside of a user's machine:
- Globus Search
- Our MLFlow model repository
- DataCite

There should be some global switch (like a GARDEN_ENV environment variable) that controls which version of these external resources we should write to. Most obviously, we will need a separate Globus Search index. There will need to be a little research done on how we handle the other two. Do we need an entire second instance of MLFlow or can we somehow tag dev models? Similarly, is there a good way to cordon off the dev vs real DOIs from DataCite?

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a developer testing Garden, when I set GARDEN_ENV=dev, then I know I can test to my heart's delight without polluting the production search index.

OwenPriceSkelly:
closes #154 and its children #282 and #283. I'm leaving #284 as open but blocked until we're ready to actually switch over to production datacite 

## Overview

This PR is the last of the productionizing PRs for now. It cleans up a handful `GARDEN_*` constants that weren't namespaced under the `garden_ai.constants.GardenConstants` class, which is also now responsible for reading the `GARDEN_ENV` environment variable and setting constants appropriately. 

## Discussion

The default behavior is still to hit the dev resources. The only way to hit prod resources from the CLI is to set `GARDEN_ENV=prod`. 

Also, it's still impossible to mint true dois programmatically (since we're leaving #284 open). For any "production seedlings" we might want to bless with a real DOI in the meantime, I think the easiest thing to do is to mint one by hand and hardcode it in the constructor `client.create_garden(..., doi='10.bespoke/doi')`

## Testing

Tested by retracing my footsteps with the BraggNN seedling against the dev environment, and by sanity checking CLI commands with and without the environment variable set. 

## Documentation

It doesn't, but probably should. I'll add a note in the developer docs 

----
:books: Documentation preview :books:: https://garden-ai--297.org.readthedocs.build/en/297/

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,3 +3,4 @@ max-line-length = 150
 extend-ignore = E203
 exclude =
     fixtures
+    demos
--- a/garden_ai/app/garden.py
+++ b/garden_ai/app/garden.py
@@ -11,9 +11,9 @@ from rich.prompt import Prompt
 from garden_ai import local_data
 from garden_ai.client import GardenClient
 from garden_ai.globus_search.garden_search import (
-    GARDEN_INDEX_UUID,
     RemoteGardenException,
 )
+from garden_ai.constants import GardenConstants
 from garden_ai.gardens import Garden
 from garden_ai.pipelines import RegisteredPipeline
 from garden_ai.app.console import console, get_local_garden_rich_table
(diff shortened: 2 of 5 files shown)


.flake8
OwenPriceSkelly, line 1: this is just for me. my local demo scripts live in a `demos` subdirectory

WillEngler (approved):
Nice!

PR #302, opened 2023-10-11
Multiple publication through prototype

### LINKED ISSUE #299: Register multiple pipelines from a notebook in a single publish command
The first draft of the publishing prototype finds the first function marked as a Garden pipeline in a `session.pkl` and runs it. This is a good start! But we would like to be able to register multiple pipelines per notebook*. 

*(I'm saying notebook here as a shorthand. This should generalize to the non-notebook approach.)

# Assumptions:
There are a lot of ways you can approach this and I won't be too prescriptive as we're still solidifying how all of this works. But here's one way you can go about it.

1. Multiple pipelines per notebook implies that just one external pipeline metadata json file won't be enough. I suggest adding metadata you need to the pipeline decorator. (This might mean that this ticket intersects with #294.)
2. But for this ticket I wouldn't focus too hard on finding the "right" way to associate pipeline metadata. I'm most interested in the raw functionality of pushing up multiple pipelines that use the same container and session.pkl.

# Acceptance Criteria
Given I have authored a notebook with two or more `@garden_pipeline`s and planted the container, when I run `garden-ai prototype register [planted container]`, then all of the tagged pipelines are registered as separate pipelines.

isaac-darling:
Fixes #299 

## Overview

Adds more comprehensive metadata support to functions decorated with the `garden_pipeline` decorator and uses this addition to employ multiple publication/execution. (i.e. multiple uniquely executable publications in one call to `garden-ai prototype register [planted-image]`)

## Discussion

This implementation is very subjective, and open to input on all fronts.

## Testing

Did a complete test run-through with a notebook example, and, if so desired, another run can be done with a normal Python script.

## Documentation

This whole thingamajig deserves some rather serious documentation overhaul, and this bit in particular will make the structure rather different compared to without.

----
:books: Documentation preview :books:: https://garden-ai--302.org.readthedocs.build/en/302/


garden_ai/app/prototype.py
```diff
@@ -246,64 +259,89 @@ def _funcx_invoke_pipeline(*args, **kwargs):
...
         f"docker.io/{image}", "docker"
     )
-    meta["container_id"] = container_id
+    container_uuid = UUID(container_id)
+    # print(f"Your container has been registered with UUID: {container_id}")
+
+    pipeline_metas = []
+    total_meta = _extract(image)
+    for key, meta in total_meta.items():
+        if "." in key:  # ignore connectors metadata
```
WillEngler, line 308: How does the period indicate that it's connectors?

WillEngler (approved):
This looks good! We're figuring out how it's all gonna plug together in real time. This is a good step towards the full experience.

garden_ai/app/prototype.py
isaac-darling: The naming scheme for the keys is laid out in `_extract_metadata_from_planted_container`, basically its the `pipeline.short_name` for the pipeline metadata (which I don't believe can contain periods since it must be an identifier) and `f"{pipeline.short_name}.connectors"` for the pipeline's connectors.

WillEngler: Gotcha, thanks!

PR #305, opened 2023-10-12
Clean out Pipeline 

### LINKED ISSUE #304: Make Pipeline object usable from notebook publication setting and clean up cruft while we're at it
Big goal: get the Pipeline class to a place where it can plug in to the new workflow without kludging by adding no-op steps and stuff like that. If possible, consolidate down to one Pipeline class. Maybe just "PipelineMetadata" or similar.

As a corollary, wipe out CLI tasks and internal functionality that depends on the deleted classes and objects. (eg, the bulk of the `pipeline` subcommands.)

# Assumptions:
1. This will break more of the old flow functionality. Basically you will no longer be able to publish pipelines the old way after this ticket lands.
2.

# Acceptance Criteria
Given I am a user adding a PipelineMetadata object to my pipeline decorator, when I fill out its fields, then I don't need to add wonky things like Steps to that object to get it working.

WillEngler:
First part of #304 

## Overview

Removes `Step`s and `Pipeline`s, leaving the just-metadata `RegisteredPipeline`. (We may want to rename that in a further refactor.)

## Discussion

The RegisteredPipeline class stays largely the same, but has a backwards-incompatible change. It no longer has a `steps` field, which had been required previously.

## Testing

In addition to getting the unit tests to pass, I smoke tested with the prototype publication flow to make sure that I could still publish a pipeline with the prototype.

## Documentation

This updates the docs by largely removing now-outdated information on how to create and publish pipelines. Plus anything about steps.

----
:books: Documentation preview :books:: https://garden-ai--305.org.readthedocs.build/en/305/


docs/architecture_overview.md
```diff
@@ -3,98 +3,24 @@ alias: Architecture Overview
...
-#  looking_glass_pipeline.py
-...
-client = GardenClient()
-looking_glass_pipeline: Pipeline = client.create_pipeline(
-    title="Looking Glass Pipeline",
-    steps=(preprocessing, run_inference),  # functions decorated above with @step
-    requirements_file="/path/to/requirements.txt",
-    authors=["Dee, Tweedle", "et al."],
-    contributors=["Dum, Tweedle"],
-    description="Makes you feel a little sorry for the poor oysters",
```
OwenPriceSkelly, line 68: 🤧 O Oysters, you've had a pleasant run 🫡

garden_ai/gardens.py
```diff
@@ -226,7 +226,7 @@ def _sync_author_metadata(self):
         known_authors = set(self.authors) - known_contributors
 
         for pipeline in self._pipeline_cache:
-            new_contributors = set(pipeline.authors) | set(pipeline.contributors)
+            new_contributors = set(pipeline.authors)
```
OwenPriceSkelly, line 229: kinda fun that this is the only change to `gardens.py`

garden_ai/pipelines.py
```diff
@@ -88,324 +73,37 @@ class Paper(BaseModel):
...
 class RegisteredPipeline(BaseModel):
-    """Metadata of a completed and registered `Pipeline` object. Can be added to a Garden and execute on a remote Globus Compute endpoint.
-
-    Unlike `Pipelines`, `RegisteredPipelines` can be described completely by JSON (since they don't need direct references to functions).
+    """Metadata of a completed and registered `Pipeline` object.
+    Can be added to a Garden and execute on a remote Globus Compute endpoint.
 
-    Note:
-        Attributes are nearly identical to [Pipeline][garden_ai.pipelines.Pipeline], with a few exceptions.
+    `RegisteredPipelines` can be described completely by JSON
```
OwenPriceSkelly, line 80: I'm wondering if `RegisteredPipeline` should just become `PipelineMetadata`? Especially since it's going to serve a basically similar role to `ModelMetadata` in the new publishing workflow

garden_ai/utils/misc.py
```diff
@@ -21,80 +16,6 @@
...
-    # to g's input types and f's return type
-    f_of_g.__signature__ = Signature(  # type: ignore
-        parameters=list(g_sig.parameters.values()),
-        return_annotation=f_sig.return_annotation,
-    )
-    f_of_g.__name__ = f.__name__ + "_COMPOSED_WITH_" + g.__name__
-    return f_of_g
-
-
 def garden_json_encoder(obj):
```
OwenPriceSkelly, line 19: I think this one can go too! Since we're not trying to serialize function types to json anymore
```suggestion

```

garden_ai/utils/misc.py
```diff
@@ -21,80 +16,6 @@
 issubtype = beartype.door.is_subhint
```
OwenPriceSkelly, line 16: `beartype` stuff can go too, I think this `issubtype` alias is the only reason it didn't get flagged as an unused import

pyproject.toml
OwenPriceSkelly, line 1: github won't let me line-comment but we can drop the beartype dependency here too

OwenPriceSkelly (approved):
doing the lord's work here. LGTM!

garden_ai/pipelines.py
WillEngler: I agree, I think it should! I just think there was enough going on focusing on the deletions here without getting into renaming and refactoring. But I think I'll probably go for the rename in my next PR.

WillEngler:
Thanks for the review! Gonna get rid of the extra stuff you pointed out and pull it in.

garden_ai/utils/misc.py
WillEngler: It looks like we hit

```
E       TypeError: Object of type UUID is not JSON serializable
```

in the unit tests if I try to take this out of the local DB operations. So this accidentally got UUID serialization working. Not gonna tackle it here but let's try and see if we can mop it up later.

PR #310, opened 2023-10-19
add metadata class

### LINKED ISSUE #304: Make Pipeline object usable from notebook publication setting and clean up cruft while we're at it

isaac-darling:
Fixes #304 

## Overview

Add a `PipelineMetadata` class used for representing and enforcing the required metadata for registering pipelines.

## Discussion

Included in this code is a wholly commented out classmethod for generating a `RegisteredPipeline` from an instance of the new class. I wrote it, but then decided it was probably poor design to expose to the user (who would not need it for any reason I could discern). If the reviewer disagrees it can be uncommented, otherwise it will be removed before merging.

## Testing

One of the changes here is to make every field of the `RegisteredPipeline` required, which breaks a substantial number of tests. This, of course, remains to be remedied.

Not sure what could go wrong with the new class, but if the structure is good then I can make a couple just to prove it works. More complete testing will occur as I do full run-throughs in my testing of multiple publication.

----
:books: Documentation preview :books:: https://garden-ai--310.org.readthedocs.build/en/310/

### DIFF
--- a/garden_ai/pipelines.py
+++ b/garden_ai/pipelines.py
@@ -195,12 +195,7 @@ class RegisteredPipeline(BaseModel):
         data = json.loads(record)
 
         if doi is not None:
-            data = {**data, "doi": doi}
-        elif "doi" not in data:
-            raise ValueError(
-                "Missing required field: doi. Please add a doi to your `PipelineMetadata` "
-                "or pass it to this function."
-            )
+            data["doi"] = doi
 
         return cls(func_uuid=func_id, container_uuid=container_id, **data) """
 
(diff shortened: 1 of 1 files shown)


WillEngler (approved):
Let's pull it in!

PR #315, opened 2023-10-23
Make container_uuid a mandatory field 

### LINKED ISSUE #311: Make container_uuid a required field in PipelineMetadata
Cleanup following up on a little discovery from #310. The container_uuid field on RegisteredPipeline has been Optional, but it should be required. This requires some cleanup/editing of our test fixtures.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Fixes #311 

## Overview

We want to make sure that all the fields we need for RegisteredPipeline to work in practice are marked as mandatory in its Pydantic definition. The last one to clean up was container_uuid.

## Discussion

Per comments that Owen left on my last PR, I also took out the `garden_json_encoder`. It could be nice to go even further and replace our use of the UUID class in Pydantic fields and just use a string field with a UUID validator. Then we wouldn't have to specify the Pydantic encoder anywhere. But this is a nice little improvement for now :) 

## Testing

I just relied on the unit tests for this.

## Documentation

Nothing big docs-wise here.

----
:books: Documentation preview :books:: https://garden-ai--315.org.readthedocs.build/en/315/

### DIFF
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -13,8 +13,9 @@ from pydantic import (
     root_validator,
     validator,
 )
+from pydantic.json import pydantic_encoder
 
-from garden_ai.utils.misc import JSON, garden_json_encoder
+from garden_ai.utils.misc import JSON
 
 from .datacite import (
     Contributor,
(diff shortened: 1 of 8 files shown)


OwenPriceSkelly (approved):
lgtm!

Issue #292, opened 2023-10-05
Add options for base image


kjschmidt913:
_Short description_
As a publisher, I want to have options for the docker base image I'm creating for my model/notebook environment so I can have my preferred python version and don't have to install common libraries every time.
# Assumptions:
1. Users can choose from a curated list of images
2. The images are described/documented on our website and documentation
3. There is a default image if the user doesn't choose one

# Acceptance Criteria
Given I have Garden installed, when I type `garden-ai prototype notebook` then I get a notebook based on the default image.
Given I have Garden installed, when I type `garden-ai prototype notebook —base-image=3.10` then I get a notebook based on the selected image.
Given I have Garden installed, when I type `garden-ai prototype notebook —base-image=foo-bar` then I get a helpful error message.

WillEngler:
Suggestions for a first pass at this:
- 3 variations with different Pythons: 3.9.x, 3.10.x, 3.11.x. They just have the basic stuff that Isaac put in the first draft of the base container in containerize.py (ie, git, jupyter, dill, etc.)
- The PR should include the Dockerfile(s) you use to make the containers and any script for building them
- You can host the containers on Dockerhub or similar under your email. We'll worry about making an official Garden or Accelerate account or hosting on our AWS ECR later. 

Suggestions for a second pass:
- Pytorch is the most popular deep learning library and takes the longest to install. Make a "matrix" of containers with different Python versions and Torch installed/not installed. So people deploying torch models can just build on a torch base container. But people who don't need the bloat don't need it.
- Make an "everything bagel" container, similar to Google Colab's default environment. To make it easy for people to get started, Colab has lots of reasonable default libraries for data science installed when you boot it up. Make our default container have Python 3.10, sklearn, torch, pandas, matplotlib, and other libraries that you deem essential for the everything bagel.

In all cases, for now do not bake in `garden-ai`. We're iterating quickly on the SDK and it'll be better to install it fresh every time at the moment.

WillEngler:
Max got in #317 which has us most of the way there on this. We need one follow-up PR with some CLI and error handling polish.

Currently, if you select a base image not in the list, Garden will attempt to pull that image from Docker Hub. We should instead show an error message and list the available images. (If we want to allow arbitrary images for internal testing or whatever, we can make another hidden option.)

Also, the help text for the `start` command should show an example in --base-image, as well as reference the command to list all the images.

PR #317, opened 2023-10-26
Garden base images

MaxTuecke:
`Adds #292`

Adds matrix of docker images that can be pulled and used in `notebook run`
Adds `notebook list-premade-images` command
Adds status message when pulling images.

New base images available:
'3.8-base', '3.9-base', '3.10-base', '3.11-base', '3.8-sklearn', '3.9-sklearn', '3.10-sklearn', '3.11-sklearn', '3.8-tensorflow', '3.9-tensorflow', '3.10-tensorflow', '3.11-tensorflow', '3.8-torch', '3.9-torch', '3.10-torch', '3.11-torch', '3.10-all-extras'

----
:books: Documentation preview :books:: https://garden-ai--317.org.readthedocs.build/en/317/


WillEngler (approved):
Nice! This is a really good first stab at this with a strong matrix right off the bat.

I do think we'll want to do a second pass for maintainability. For example, there is a lot repeated between the Dockerfiles and updating them will probably be fiddly. It would be cool if we could do codegen to make them so that, say, we could inject a different Jupyter token without going and editing all the files. And as a team we also need to back up and get our one shared Dockerhub account figured out, haha.

But! In the spirit of prototyping, this is just what the doctor ordered. Thanks!

Issue #314, opened 2023-10-23
garden-ai notebook publish


WillEngler:
As a Garden publisher, I want one command to go from my notebook to having my models discoverable and executable.

Make a `garden-ai notebook publish [notebook path]` command. This should build an image on top of the base image the user selected initially, or let the user override that with a `base-image` option. It should register all the pipeline functions present in the notebook, associate the pipelines with the specified gardens, and publish all the affected gardens.

# Assumptions:
1. The publisher must have already created a Garden. 
2. the publisher must specify a garden doi in every @garden_pipeline decorator

# Acceptance Criteria
Given I have finished my publication notebook, when I run `garden notebook publish [my-notebook]`, then all the pipelines in my notebook are now available for execution.

PR #322, opened 2023-10-30
314 notebook publish

OwenPriceSkelly:
Implements #314 🎉

## Overview
This implements the `garden-ai notebook publish [path/to/notebook]` command. 

- Takes a notebook with at least one `@garden_pipeline`-decorated function, a base image (or last used for that notebook), and an image repo (or last one specified by user)
- builds and pushes the "planted" image containing the original notebook, notebook-as-script, session.pkl, and metadata.json in the `/garden` directory 
- registers that new image and the respective "load session and call the function" functions with globus compute. 
  - These are now simple enough functions that we can register them as plain source code, which is cool 
- creates the new RegisteredPipeline objects for each decorated function and saves to local data 
  - if a garden_doi was specified in the decorator, the pipeline is automatically added to that garden and the updated garden metadata is written to local data and republished to the search index
  - if not, you can still use the CLI `garden-ai garden add-pipeline` to add the pipeline to a garden at a later point (instead of always requiring a garden doi)

## Discussion
The important new machinery lives in a few different places:
- the publish command itself in `garden_ai.app.notebook` (duh)
- the core docker sdk helpers live in `garden_ai.containers`
- the code to register the pipelines and re-publish gardens is a big `GardenClient` method 
- the script we append to the user's code is `garden_ai.scripts.save_session_and_metadata`
- the functions we actually register with globus compute are generated by `utils._meta.make_function_to_register` 
 
## Testing

chatgpt and I wrote some unit tests for the `garden_ai.containers` functions, but I didn't care enough to deal with mocking everything I'd need to for the big `GardenClient._register_and_publish_from_user_image` method. 

Aside from those, I tested this as I went with a very basic notebook that did some pip installs and defined a dummy pipeline, and confirmed that the whole process does indeed make the pipeline executable on the dlhub endpoint via the published garden that I specified. 

## Documentation

I was as anal about comments/docstrings as I usually am, but didn't add anything specific to the docs. I think that we'll probably want a whole new tutorial/overview section and that feels like its own story to me, and I didn't think it was worth blocking this PR on that.  

----
:books: Documentation preview :books:: https://garden-ai--322.org.readthedocs.build/en/322/


garden_ai/containers.py
```diff
@@ -53,3 +60,149 @@ def start_container_with_notebook(
...
+                    print(chunk["stream"].strip())
+            if "aux" in chunk and "ID" in chunk["aux"]:
+                image_id = chunk["aux"]["ID"]
+                image = client.images.get(image_id)
+            if "error" in chunk or "errorDetail" in chunk:
+                error_message = chunk.get(
+                    "error", chunk.get("errorDetail", {}).get("message", "")
+                )
+                print("Build failed:", error_message)
+                raise docker.errors.BuildError(reason=error_message, build_log=stream)
```
WillEngler, line 149: This looks thorough and like it took some experimentation. Nice stuff

garden_ai/containers.py
```diff
@@ -53,3 +60,149 @@ def start_container_with_notebook(
...
+        The full Docker Hub location of the pushed image (e.g. "docker.io/username/myrepo:tag" )
+    """
+    # Tag the image with the new repository name
+    image.tag(repo_name, tag=tag)
+
+    # push the image to the new repository and stream logs
+    push_logs = client.images.push(repo_name, tag=tag, stream=True, decode=True)
+    for log in push_logs:
+        if not print_logs:
+            continue
```
WillEngler, line 201: Understanding check: This weird control flow is because you have to drain the generator that you get because `stream=True`, right? I wonder if we'd be better off splitting out the print/no-print cases to avoid the `continue`, but idk if that would be more readable

garden_ai/pipelines.py
```diff
@@ -157,30 +158,24 @@ def __call__(
...
+    #     details = f"<p>Authors: {', '.join(self.authors)}<br>DOI: {self.doi}</p>"
+    #     optional = "<h3>Additional data</h3>" + tabulate(
+    #         [
+    #             (field, val)
+    #             for field, val in self.dict().items()
+    #             if field not in ("title", "authors", "doi", "steps") and val
+    #         ],
+    #         tablefmt="html",
+    #     )
+    #     return style + title + details + optional
```
WillEngler, line 178: Should we just delete this and the imports that are unused now?

garden_ai/scripts/save_session_and_metadata.py
```diff
@@ -0,0 +1,44 @@
...
+        total_meta[key_name] = marked._pipeline_meta
+        # TODO add these as regular fields in PipelineMetadata/ RegisteredPipeline?
+        total_meta[connector_key] = [
+            connector.metadata for connector in marked._model_connectors
+        ]
+        if marked._garden_doi:
+            total_meta[doi_key] = marked._garden_doi
+
+    with open("metadata.json", "w+") as fout:
+        json.dump(total_meta, fout, default=pydantic_encoder)
```
WillEngler, line 44: I really prefer this metadata.json approach to my old approach haha.

garden_ai/utils/_meta.py
```diff
@@ -27,5 +28,61 @@ def redef_in_main(obj):
...
+        DillCodeTextInspect strategy, not DillCode.
+    """
+    source = f"""
+def call_{func_name}(*args, **kwargs):
+    import __main__
+    import dill
+
+    dill.load_session("/garden/session.pkl")
+
+    return __main__.{func_name}(*args, **kwargs)
```
WillEngler, line 82: 4 lines, let's gooo

garden_ai/utils/_meta.py
```diff
@@ -27,5 +28,61 @@ def redef_in_main(obj):
...
+    getlines = linecache.getlines
+
+    def monkey_patch(filename, module_globals=None):
+        if filename == "<string>":
+            return source.splitlines(keepends=True)
+        else:
+            return getlines(filename, module_globals)
+
+    # can now use inspect.getsource() on the result of exec() here
+    linecache.getlines = monkey_patch
```
WillEngler, line 53: I think this leaves the monkey patch in place after this function has finished. (No cleanup `linecache.getlines = getlines` at the end.) Is that intentional?

garden_ai/app/notebook.py
```diff
@@ -128,318 +129,77 @@ def _register_container_sigint_handler(container: docker.models.containers.Conta
...
-    print(
-        f"Successfully registered your new pipeline(s) with doi(s): {pipeline_name_to_doi}"
+    if image is None:
+        typer.echo("Failed to build image.")
+        raise typer.Exit(1)
+    typer.echo(f"Built image: {image}")
+
+    # generate tag and and push image to dockerhub
+    timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
+    image_tag = f"{notebook_path.stem}-{timestamp}"
```
WillEngler, line 197: Really appreciate these tagging changes

WillEngler (approved):
Left some questions/comments but nothing that would block this imo. Hats off!!!!

garden_ai/containers.py
OwenPriceSkelly: yeah no agreed, reads better without the continue

garden_ai/pipelines.py
OwenPriceSkelly: Oh actually I left those commented out to remind myself to see if I could figure out how to keep it and keep dill happy once I had something working. 

That might as well be its own small issue, I'll open than and clean this up

garden_ai/utils/_meta.py
OwenPriceSkelly: yeah, unfortunately. the monkeypatch needed to stay around long enough to still apply to globus compute's call to `getsource` in the serialization strategy

OwenPriceSkelly: I think morally it's ok here because it's only going to impact cases that would otherwise have been a straight up error from `getlines` but I don't love it

PR #324, opened 2023-11-01
313 notebook debug

### LINKED ISSUE #313: garden-ai notebook debug
As a Garden publisher power user, I want a way to test the exact execution context my pipeline will run in. 

`garden-ai notebook debug` should put me in a context (likely a notebook) where I can test the built container and deserialized notebook session. 

We've tossed around some ideas on how to communicate this visually to the user. Maybe we have one big cell with everything the user wrote commented out. We might also be able to "gray out" cells by adding this to cell metadata: `{ "trusted": true, "editable": false, "deletable": false }` I'm open to whatever works.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

isaac-darling:
Fixes #313 

## Overview

Adds the debug command, with some small changes to (Owen's?) new Python docker code.

## Testing

I ran the command successfully locally.

----
:books: Documentation preview :books:: https://garden-ai--324.org.readthedocs.build/en/324/


isaac-darling:
My change to not force the client to pull an image breaks a new test, which of those things is to be changed is up to the reviewer.

garden_ai/containers.py
```diff
@@ -30,14 +33,24 @@ def start_container_with_notebook(
...
     - The token for accessing the notebook is still the JUPYTER_TOKEN variable.
     """
-    with console.status(f"[bold green] Pulling image: {base_image}"):
-        client.images.pull(base_image, platform="linux/x86_64")
+    # first check if the image is present, otherwise attempt to pull
+    try:
+        client.images.get(base_image)
+    except docker.errors.ImageNotFound:
+        with console.status(f"[bold green] Pulling image: {base_image}"):
+            client.images.pull(base_image, platform="linux/x86_64")
```
WillEngler, line 41: This seems like a sensible change to me. @OwenPriceSkelly is more in the thick of it so I'll get his take for a second opinion

garden_ai/app/notebook.py
```diff
@@ -266,8 +266,34 @@ def debug(
...
+    )
+
+    # stream logs from the container
+    for line in container.logs(stream=True):
+        print(line.decode("utf-8"), end="")
+
+    # block until the container finishes
+    container.wait()
+    typer.echo("Notebook has stopped.")
+    return
```
WillEngler, line 296: Nice!

WillEngler (commented):
LGTM, but I'll punt to Owen for a look at the behavior that breaks the test. (Makes sense to me, but want to be sure there isn't something non-obvious he had in mind.)

garden_ai/containers.py
OwenPriceSkelly: I thought this was already the default behavior of a naive docker pull but I could be mistaken! what kind of case would this behave differently? if it's equivalent I'd just as soon leave it

isaac-darling: The impetus was that before I implemented the change, I got an error that I was not permitted to pull from `gardenai`, since the default image name is `gardenai/base` and of course I don't have such a permission.

OwenPriceSkelly: I think that's because `gardenai/base` doesn't exist yet, not a permission thing -- should have defaulted to `gardenai/test` which is hopefully public, if you still can't pull that one I need to fix that

isaac-darling: I guess it then becomes this: we need our default image names to be things that theoretically could be pulled (which feels counterintuitive to me, since we are doing so knowing that the image has no meaning and will never actually be pulled), or we only attempt to pull when the image is not already present.
Reading what I wrote, my bias is very clear, but I hope it doesn't take away from the binary.

OwenPriceSkelly:
no major gripes but let's hold off on merging this one until my current PR is in -- I might split off what I have so far into its own PR so you can build on top of it. specifically, I've got the new helper to build the baked/planted/etc container done, and I think the debug command should put users in that container rather than a base image

OwenPriceSkelly (changes requested):
This is a great start! Once I got into the debug notebook it seemed like everything was right, my only hard blocker is the error I ran into with the image not being tagged yet. Other comments are just misc enhancements that should be fast. 

If building a debug notebook that displays the user's source seems like it's a significant lift then it can be its own story and unblock this one, but I have a hunch that `nbformat` has the machinery to make it pretty straightforward

garden_ai/app/notebook.py
```diff
@@ -137,7 +137,61 @@ def handler(signal, frame):
...
+    """
+    docker_client = docker.from_env()
+    base_image = _get_notebook_base_image(path) or "gardenai/test:latest"
+    image = build_notebook_session_image(
+        docker_client, path, base_image, print_logs=False
+    )
+    if image is None:
+        typer.echo("Failed to build image.")
+        raise typer.Exit(1)
+    image_name = str(image.tags[0])  # str used to guarantee type-check
```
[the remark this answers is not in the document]
OwenPriceSkelly: best way to tweak it might be to give `start_container_with_notebook` a `pull=True` kwarg and call it like 
`container = start_container_with_notebook(..., base_image=image.id, mount=False, pull=False)`, since images always have an `image.id` even if they're not yet tagged

OwenPriceSkelly (approved):
lgtm!🌱

Issue #294, opened 2023-10-05
Pipeline Metadata Design Update


kjschmidt913:
_Short description_
As a consumer, I want there to be information about the pipeline so I can understand if it meets my needs.
As a publisher, I want to attach the metadata for my pipeline to my notebook so consumers can understand if it meets their needs.

# Assumptions:
1. Deliverable will be an update to the existing pipeline pydantic class
2. Use frontend designs in Figma for additional metadata fields
3. The schema is documented in the Garden documentation website

# Acceptance Criteria

Given I am on the Garden docs, when I click on `metadata schema`, then I see the correct metadata schema in an easy to understand format.

WillEngler:
Superseded by #314 and new metadata design-specific ticket that we are about to write

WillEngler:
Merged #330 and #326 with updates on models, datasets, and pipelines that I'd sketched out in the FE mocks (https://github.com/Garden-AI/garden-frontend/pull/58). Here's the current readthedocs for the Pipeline class: https://garden-ai.readthedocs.io/en/latest/Pipelines/

There's gonna be follow-up work to pipe all the data end to end, but I think we've hit the mark on updating the metadata design.

WillEngler:
Had some good follow-up conversations where we hashed out more details on the metadata. Basically we're good with the format worked out in the PRs I got in, plus the re-addition of steps. Going to mark this design phase of work as done. There are now more targeted tickets for getting the metadata flowing end to end: #334, #339, https://github.com/Garden-AI/garden-frontend/issues/61

PR #326, opened 2023-11-06
Update model and dataset metadata

WillEngler:
Part one of #294 

## Overview

This PR updates metadata related to models and datasets to match the new way we're thinking about models. (ie, we're linking to models in external model repositories. We don't need to store information on how to serialize/deserialize them ourselves.)

## Discussion

This PR makes it so that models are no longer top-level entities in our datastore. A lot of deletions follow from that. Now model metadata is stored directly in Pipeline entities rather than having a key-value lookup relationship.

## Testing

I just relied on the unit tests here.

## Documentation

The change where Pipelines now directly contain Model metadata (not just names) is reflected in the autogenerated pipeline metadata docs


OwenPriceSkelly (approved):
lgtm!

PR #346, opened 2023-11-15
Add steps to the new publishing flow with @garden_step decorator

### LINKED ISSUE #334: Add @garden_step decorator and propagate steps into metadata
Bring back [The Steps](https://open.spotify.com/track/5X2RisOZS9i5MsR8EVw58p?si=fb96f8d1e77448d6). Add a `@garden_step` decorator that notebook authors can use to optionally tag functions. These functions will be considered Steps for their pipeline. 

# Assumptions:
1. We aren't going back to having Steps be stored on Pipeline objects as Python Callables. The important thing here is capturing the text of the function so we can include it and send it along as metadata so the UI can display it.

# Acceptance Criteria
Given I make a pipeline without any @garden_step decorated functions, when I publish the pipeline, then it will show in the UI with one step. (Just the @garden_pipeline)

Given I make a pipeline with @garden_step decorated functions, when I publish the pipeline, then it will show in a UI with all the tagged step functions plus a step for the @garden_pipeline

# Gametime Decisions

When there are multiple `@garden_pipeline`s in a notebook, should we require that the user explicitly list which steps apply to which? My instinct for this draft is to say that you can optionally pass a list of steps to a @garden_pipeline decorator but we don't make you. By default we assume every step in a notebook applies to every pipeline, unless the user tells us otherwise.

WillEngler:
Closes #334 

## Overview

This PR re-introduces steps to the new publishing flow. Users can decorate functions in their notebook with `@garden_step` and that will expose the function as a step in their pipeline's metadata so that the UI can feature it. Pipeline functions (the ones with `@garden_pipeline`) are also steps by default.

For now there is no way to specify the order of your steps. The order is always the order they are defined in your notebook, with the pipeline function always the final step. If there is more than one pipeline in a notebook, all the steps apply to both pipelines. If this idea takes off we'll probably want to let users specify which apply to which, but the intention here is to cover the simple case first so we can see it working end to end and go from there.

## Discussion

I made some light changes to how the decorators store data on the functions they decorate. To make names and attributes easier to keep track of, I tried to get it down to just sticking a Step or Pipeline object on decorated functions. To do this I added private Pydantic attributes to PipelineMetadata that won't show up in the documentation or the serialized JSON.

## Testing

I added a simple unit test for the new decorator. But the bulk of the testing came from publishing a bunch of pipelines and confirming that steps were serializing how I expected. (Including for multi-pipeline notebooks.)

## Documentation

No docs on this yet

----
:books: Documentation preview :books:: https://garden-ai--346.org.readthedocs.build/en/346/

### DIFF
--- a/garden_ai/pipelines.py
+++ b/garden_ai/pipelines.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
+import inspect
 import logging
 from datetime import datetime
 from typing import Any, List, Optional, Union
 from uuid import UUID
 
 import globus_compute_sdk  # type: ignore
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, PrivateAttr
 
-# from garden_ai.app.console import console
 from garden_ai.constants import GardenConstants
 from garden_ai.datacite import (
     Creator,
@@ -73,6 +73,24 @@ class Paper(BaseModel):
     citation: Optional[str] = Field(None)
 
 
+class Step(BaseModel):
+    """
+    The `Step` class represents a key function in a pipeline that a publisher wants to highlight.
+
+    Attributes:
+        function_name (str):
+            The name of the step function.
+        function_text (str):
+            The full Python code that makes up the function.
+        description (str):
+            An optional string describing the function.
+    """
+
+    function_name: str = Field(...)
+    function_text: str = Field(...)
+    description: Optional[str] = Field(None)
+
+
 class PipelineMetadata(BaseModel):
     """Metadata for a pipeline prior to its registration. \
     Passed to the `garden_pipeline` decorator during the registration process.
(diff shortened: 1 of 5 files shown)


garden_ai/pipelines.py
```diff
@@ -207,9 +229,26 @@ def garden_pipeline(
...
+
+    return decorate
+
+
+def garden_step(description: str = None):
+    def decorate(func):
+        func._garden_step = Step(
+            function_text=inspect.getsource(func),
+            function_name=func.__name__,
+            description=description,
```
OwenPriceSkelly, line 250:
```suggestion
            description=description or func.__doc__,
```
might as well use a docstring for description if func has one

OwenPriceSkelly: maybe take or leave it, since we would already have a docstring show up in `function_text`, too

OwenPriceSkelly (approved):
LGTM! seems like a very sane approach to steps right now before we know what they're really going to look like

garden_ai/pipelines.py
WillEngler: I'm very open to this. Gonna leave it as is for now but we'll probably have another pass at this once we see it end to end in the UI

Issue #338, opened 2023-11-10
Figure out a recommended Tensorflow workflow for Apple Silicon users


WillEngler:
Booting up one of the tensorflow base images on my Macbook and then running `import tensorflow as tf` grinds the notebook session to a halt. This is not super surprising. This is why there is a specific `tensorflow-macos` package. But this does break the idea of Garden letting you test on the _exact_ container and libraries locally, then shipping it off.

What to do? Is there a workaround here? (Test locally on the macos variant, send off a remote build, and then pray?) Do we just recommend people use a remote workstation for this? Not sure yet.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

PR #353, opened 2023-11-28
340 docs update

### LINKED ISSUE #301: 🪲Bug Hunt🪲: dill goes to jupyter. what gets stupyder?
Our new publication workflow takes an "everything but the kitchen sink" approach to serializing user code, which _significantly_ reduces the hoops our sdk and the users need to jump through in order to publish their work. The tradeoff is that the new approach leans heavily on `dill`'s ability to reliably save a python interpreter session to a file, and opens us up to far more edge cases. Now that we're going to try and serialize arbitrary jupyter notebooks with dill, we need to know what kinds of jupyter notebook state dill won't be able to handle as advertised. 

At a high level, this is the process:
- Given some jupyter notebook, convert it to a plain python script with `nbconvert` 
- Run that script in a fresh python interpreter (bringing user definitions, imports etc into scope)
  - Ideally, this should reach exactly the same interpreter state as running each cell in order manually would. When might this not be true?
- Save the state of that python process to a file using `dill.dump_session` 
- Later, give that file to `dill.load_session` in another fresh process in order to call one of the user-defined functions. 

## What we are *not* worried about: 
- Problems due to a mismatch between python, jupyter, dill versions or OS. Assume both sides of the serialize/deserialize process are properly containerized, just different python processes
  - NB: Any version-related bugs or incompatibilities that keep the environment fixed on both sides still count! 
  
## What we *are* worried about:
- jupyter notebook magic/state that is lost when run as a plain python script 
- notebook state that would be preserved but is un-pickleable and causes immediate problems on the serialization side 
- notebook state that _seems_ to pickle fine but causes problems when deserializing later

# Acceptance Criteria
- Description of the bug with as much relevant detail as you have
- Step-by-step instructions (or, even better, a script) for reproducing the issue.
- Bonus points for any ideas on workarounds, but not necessary 

# Deadline and Prize for Best Bug 
- Best Bug is subjective (but we have to be able to reproduce it!). More than open to argument on the basis of size, technical interest, relevance to users, fun factor, style points, etc. 
- Bug description and repro steps should be linked under this issue before **November 21** 
- the Best Bug submitted before the deadline takes home this lil guy: 
image (1)

### LINKED ISSUE #340: tutorial/first draft of docs for new publishing flow
Documentation currently has a few `<x> is currently under construction. Check back soon. 👷🏽` notes that should be fleshed out with a helpful tutorial using the new publishing flow. 

CLI commands should have docs somewhere besides `--help` 

There should also be a "limitations"/"pitfalls" page we can update as a reference with stuff that we know has the potential to cause problems or unexpected behavior for users but can't necessarily prevent

OwenPriceSkelly:
closes #340 

This includes updated user docs for installation, overview, and tutorial sections. I'm leaving the "limitations" page for another PR that will close #301 and #338

----
:books: Documentation preview :books:: https://garden-ai--353.org.readthedocs.build/en/353/

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,8 +30,9 @@ jobs:
           poetry-version: 1.2.2
       - name: Install dependencies
         run: |
-          poetry config virtualenvs.create false && \
-          poetry install --no-interaction --no-ansi --with=test
+         pip install --upgrade pip && \
+         poetry config virtualenvs.create false && \
+         poetry install --no-interaction --no-ansi --with=test
       - name: black lint
         uses: psf/black@stable
         with:
--- a/docs/architecture_overview.md
+++ b/docs/architecture_overview.md
@@ -1,60 +1,64 @@
 ---
 alias: Architecture Overview
 ---
-## Overview
+## Garden Project Conceptual Overview
 
-The Garden project is structured around two core concepts: `Gardens` and `Pipelines`. Each of these is represented by one or more classes in our SDK.
+The Garden project is structured around two "proper noun" concepts: `Garden` and `Pipeline`.
 
-`Garden` and `Pipeline` objects define the *citable* and *reproducible* end products, and enable users to discover/share scientific work.
+`Garden` and `Pipeline` objects define the *citable* and *reproducible* end products, and enable users to discover and share each other's scientific work.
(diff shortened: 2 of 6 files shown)


WillEngler (approved):
Heck yeah!!! Let's merge this

Issue #345, opened 2023-11-15
Move notebook strings out of search payloads


WillEngler:
So far we've gotten away with just plopping Jupyter notebook JSON right into our search ingest requests. It seems like this breaks for bigger notebooks. (eg, one I was testing with that's >25k characters) I'm not 100% sure this is the culprit because the message from Search is ambiguous ("Unexpected errors occurred on document index/update"). But from playing around with it, I'm pretty confident.

We should probably toss the notebook contents into S3 and put the S3 url into the search record. tbqh I'm surprised we got this far without needing to do this.

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a publisher, when I try to publish with a large-ish notebook, then it works.

WillEngler:
Subtasks:

Infra: new dev and prod S3 buckets for storing notebook ipynb files. Write access is restricted to Garden backend. Read access is fully open with CDN caching enabled.

Backend: make new `POST /jupyter-notebook` endpoint. It takes notebook JSON as its payload. It writes the notebook file to the right bucket at `{user-id}/{notebook-name}-{hash}`. It returns a publicly accessible URL that points to the notebook file.

CLI: New `--keep-notebook-output` option on `notebook publish` command. Default behavior is that we always clear the cell output before pushing the notebook up. But users can opt to keep the output if they want to. We refuse to try any notebook payload larger than 5MB. (This keeps us under API Gateway and Lambda payload limits.) The CLI now invokes the new backend endpoint at publication time. We now use the notebook URL for the `notebook` metadata fields instead of the notebook JSON contents.

Frontend: The pipeline page makes a follow-up call to the notebook URL in the pipeline metadata if/when it wants to display the notebook contents.

WillEngler:
All 5 out of 4 PRs are in :) Closing

PR #354, opened 2023-11-28
Attach notebook URL to metadata instead of full notebook contents

WillEngler:
This is the SDK/CLI portion of #345 

## Overview

This PR uses the new POST /notebook endpoint to upload the contents of the user's notebook to S3 and attach a public URL to a new `notebook_url` field on the pipeline metadata.

## Discussion

This PR is pretty by the book, but one interesting part is how we treat outputs in notebooks. For both uploading and downloading notebooks, we want to try and keep the file size sane. The most likely way that notebook file size gets out of control is `outputs`. So now we nondestructively clear the user's outputs by default before uploading. If the user has some visualization or output they really want to show, they can provide the --keep-outputs flag.

## Testing

I added new unit tests for the notebook helpers. I also tested this end to end against the backend dev instance and confirmed that `notebook_url` was getting included as expected, with a valid URL that points to the notebook contents.

## Documentation

No docs changes

----
:books: Documentation preview :books:: https://garden-ai--354.org.readthedocs.build/en/354/

### DIFF
--- a/.safety-policy.yml
+++ b/.safety-policy.yml
@@ -16,6 +16,8 @@ security: # configuration for the `safety check` command
             reason: this is a vulnerability of a dependency of mlflow. We don't use the subset of mlflow that uses this
         60841: # Vulnerability ID
             reason: this is a vulnerability of a dependency of mlflow. We don't use the subset of mlflow that uses this
+        62044:
+            reason: This doesn't affect us and Owen has resolved this on another branch.
     continue-on-vulnerability-error: False # Suppress non-zero exit codes when vulnerabilities are found. Enable this in pipelines and CI/CD processes if you want to pass builds that have vulnerabilities. We recommend you set this to False.
 alert: # configuration for the `safety alert` command
     security:
--- a/garden_ai/app/notebook.py
+++ b/garden_ai/app/notebook.py
@@ -4,6 +4,7 @@ import shutil
 import webbrowser
 from pathlib import Path
 from typing import Optional
+import json
 
 import docker  # type: ignore
 import typer
@@ -16,6 +17,7 @@ from garden_ai.containers import (
     start_container_with_notebook,
 )
 from garden_ai.local_data import _get_notebook_base_image, _put_notebook_base_image
+from garden_ai.utils.notebooks import clear_cells, is_over_size_limit
 
 logger = logging.getLogger()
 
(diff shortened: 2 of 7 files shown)


.safety-policy.yml
```diff
@@ -16,6 +16,8 @@ security: # configuration for the `safety check` command
             reason: this is a vulnerability of a dependency of mlflow. We don't use the subset of mlflow that uses this
         60841: # Vulnerability ID
             reason: this is a vulnerability of a dependency of mlflow. We don't use the subset of mlflow that uses this
+        62044:
+            reason: This doesn't affect us and Owen has resolved this on another branch.
```
OwenPriceSkelly, line 20: 😇

tests/test_utils.py
```diff
@@ -13,3 +14,53 @@ def test_extract_email_from_globus_jwt_malformed(identity_jwt):
...
+
+
+def test_clear_cells():
+    assert clear_cells(notebook_with_outputs) == cleared_notebook
+
+
+def test_is_over_size_limit():
+    assert not is_over_size_limit(cleared_notebook)
+
+    # 6MB worth of "a"s
```
OwenPriceSkelly, line 63: the kinda notebook written by the monkey with a typewriter before he works up to shakespeare

garden_ai/backend_client.py
```diff
@@ -50,3 +50,14 @@ def update_doi_on_datacite(self, payload: dict):
     def publish_garden_metadata(self, garden: PublishedGarden):
         payload = json.loads(garden.json())
         self._post("/garden-search-record", payload)
+
+    def upload_garden_metadata(
```
OwenPriceSkelly, line 54: should this be named `upload_notebook` or something instead? or are you envisioning this method eventually doing something with garden metadata, too?

garden_ai/app/notebook.py
```diff
@@ -217,6 +219,12 @@ def publish(
         ),
     ),
     verbose: bool = typer.Option(False, "-v", "--verbose"),
+    keep_outputs: bool = typer.Option(
```
OwenPriceSkelly, line 222: I like this a lot, sane default but nice to give them the option

OwenPriceSkelly (approved):
lgtm!

garden_ai/backend_client.py
WillEngler: Oof, no this is a mistake. Probably some copypasta. I'll make it something like upload_notebook, thanks for catching

tests/test_utils.py
WillEngler: or maybe the monkey has transcended shakespeare, gone back to their roots

Issue #361, opened 2023-12-06
Second draft of tutorial


WillEngler:
Writing is rewriting!

Owen banged out a good first draft of the new Garden docs. Now we should take a second pass at it.

Some things to look out for in particular:
- Does the language in the docs and the template notebooks line up and complement each other?
- We should probably add a FAQS page that addresses known limitations

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Specifically:
- What needs to be pre-installed. (Particularly Docker and the Hugging Face account)
- Where do you find your DOI after you've made your garden?

WillEngler:
Call out the fact that we have a pre-prepared model for users to take advantage of

WillEngler:
Make very clear that people should use pipx as their first resort for installing garden-ai

WillEngler:
Kick off with objectives and some conceptual context. "The goal here is to make a function that executes your model. We call that function an entrypoint ..."

WillEngler:
Bear in mind that the "typical" user is a chemistry student. Not someone we can make analogies to Lambda with. How do we make it really clear to that persona how this whole thing fits together?

WillEngler:
Re-scoping this ticket to just include the end to end tutorial experience. That alone is a pretty big lift

WillEngler:
Closed by #398

PR #365, opened 2023-12-11
entrypoint rebrand

### LINKED ISSUE #358: rebrand `Pipelines` -> `Entrypoints`
Out with the old. The objects-formerly-known-as pipelines are going to become something else, tentatively `Entrypoint`s as of Owen and KJ syncing on 12/4. 

`Pipeline`s as a concept still exist as the special case of an `Entrypoint` (or whatever) that actually _does_ consist of multiple steps in a sequence 

(will's recap copied from slack): 
> Gardens are a collection of entrypoints. Entrypoints have at least one step. If entrypoints have more than one step, we consider them pipelines.

> At a technical level, we're good to change PipelineMetadata to EntrypointMetadata on the backend. (And similar for other occurrences of "pipeline") And that will be reflected in the serialized metadata in Globus Search. (Gardens will have lists of `entrypoint`s.) The decorator in the notebooks will be `@garden_entrypoint`. Entrypoints will still have lists of steps. As now, they'll always have at least one step. (The `@garden_entrypoint` function). And as now, the publisher can add more steps with `@garden_step`.

OwenPriceSkelly:
closes #358 

## Overview

This renames `Pipeline` to `Entrypoint` everywhere in the SDK. It turned out not to be worth it to create another distinct `Pipeline` class on top of `Entrypoint`; the pipeline-vs-entrypoint distinction doesn't need to be made by backend code because the distinction is only relevant when one is being displayed. 

## Testing

Ran unit tests and sanity checked the rebranded `garden-ai entrypoint` CLI commands  

## Documentation

This does update docs but still pretty naively. I did make an effort to fix everything that became "A entrypoint" to "An entrypoint" but held off changing any awkward wording for #361  

----
:books: Documentation preview :books:: https://garden-ai--365.org.readthedocs.build/en/365/


WillEngler (approved):
image

PR #368, opened 2023-12-11
Make sure we don't publish entrypoints with incompatible dill versions

### LINKED ISSUE #357: Prevent accidental dill slippage
The DLHub endpoint assumes that pipelines have been serialized with dill 0.3.5.1. We bake that version of dill into our base images. But what if someone installs a library with a different (later) version of dill? Then remote deserialization will probably fail.

At minimum ...
- Lock the dill version in Garden's dependencies to 0.3.5.1.

And maybe ...
- Add some check or failsafe in the publication process that aborts publication if the publication process is running with the wrong version of Dill

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a Garden publisher, when I publish my notebook, then there is no way that it gets serialized with a version of Dill that the River worker can't deserialize.

WillEngler:
Fixes #357.

## Overview

- Pins dill version in poetry.toml
- Aborts publication if we see that an environment created from a notebook/script has the wrong version of dill

## Testing

I confirmed that raising the dill error works as suspected for a notebook with and without the right version of dill.

## Documentation

None

----
:books: Documentation preview :books:: https://garden-ai--368.org.readthedocs.build/en/368/


OwenPriceSkelly (approved):
lgtm!

PR #370, opened 2023-12-13
359 error if `stage()` called during publish

### LINKED ISSUE #359: warn users if they've baked weights into their image by mistake
it's pretty easy to accidentally write test code that downloads your weights from huggingface and accidentally leave it there when publishing the notebook. 

In general we won't be able to anticipate every case of this, but if users are downloading with our recommended `HFConnector` approach we should at least warn them if the `.stage()` method is called during the publication process

OwenPriceSkelly:
closes #359 

## Overview

This decorates the `HFConnector.stage` method so that we can raise an error if it's been called during an image build, but not anywhere else.

## Discussion

Originally we wanted this to be just a warning, but that turned out to be pretty poor UX -- a warning wouldn't appear at all unless running `notebook publish --verbose`; even when verbose, the warning was hard to find in the output among the various safe-to-ignore `WARNING: Running pip as the 'root' user can result ... ` messages.

I spent a while trying to come up with a clean way to provide an option for power users to do this anyway, but after hitting a couple dead-ends I think it makes more sense for those users to just not use `.stage()` than it does for me to spend much more time on this ticket.

tldr: turned out more opinionated than I'd like, but I think it's innocent until proven gripe 

## Testing

Published the tutorial notebook (with this branch installed) a few times with/-out problematic code and the `--verbose` flag 

## Documentation

no new docs 

----
📚 Documentation preview 📚: https://garden-ai--370.org.readthedocs.build/en/370/

### DIFF
--- a/garden_ai/model_connectors/hugging_face.py
+++ b/garden_ai/model_connectors/hugging_face.py
@@ -1,5 +1,6 @@
 import huggingface_hub as hfh  # type: ignore
 from garden_ai.mlmodel import ModelMetadata
+from garden_ai.utils.misc import trackcalls
 import os
 import sys
 
@@ -19,6 +20,7 @@ class HFConnector:
             model_version=self.revision,
         )
 
+    @trackcalls
     def stage(self) -> str:
         if not os.path.exists(self.local_dir):
             os.mkdir(self.local_dir)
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
> I think it makes more sense for those users to just not use .stage() than it does for me to spend much more time on this ticket.

Love it, ship it 🚢

Issue #362, opened 2023-12-07
Display and "save" the base image a notebook uses in a notebook cell


WillEngler:
We have a little first draft system for pairing a notebook with a base image that looks like this ...

1. User says `garden-ai notebook start foo.ipynb --base-image=3.10-torch`
2. We save a mapping from `current/directory/foo.ipynb` to `3.10-torch` in the user's local data
3. Then the user can say `garden-ai notebook start foo.ipynb` and we open up it up again in the right image.

This is nice as a quick first pass but ofc it's pretty fragile. If the user `mv`s the notebook to a different directory they'll be prompted again to specify the base image. 

I suggest that when making a new notebook from a template we create a first cell that presents (in a nice user-readable format) what base image this notebook uses. If the user decides to change it later, that is well and good! They can just edit the cell.

This has a few benefits
- We don't need a local DB mapping anymore
- The user can move the notebook around or even send it to another machine and the notebook keeps its own metadata. (Especially useful if we're debugging someone else's notebook)

# Assumptions:
1.
5.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Now that we let users specify a --requirements file, that's something we can also think about in this ticket. The base image we provide and the requirements a user has specified feel pretty linked in the user's mind. So we'll want to think about those holistically.

PR #373, opened 2023-12-21
336 pre install garden in base images

### LINKED ISSUE #336: Pipeline authors should not need to !pip install garden-ai in their notebooks
One of the big gotchas in the new publication flow is that you need to !pip install garden-ai yourself in the notebook. That's surprising and it would be good if users didn't need to do it. 

We suspect the least complex way to fix this is to rebuild all of our images when we cut a release with the new garden version. 

# Assumptions:
1. First, whoever picks up this ticket should do a little designing and think through pros and cons of different options.
2.

# Acceptance Criteria
Given I am a Garden publisher, when I start a fresh notebook with `garden-ai notebook start`, then I can immediately `import` stuff from garden-ai

OwenPriceSkelly:
closes #336, plus addresses Logan's gripe about including requirements.txt or conda environment.yml files. 
## Overview

The new functionality is all in `build_image_with_dependencies`, and all of the `garden-ai notebook` commands now accept an optional `--requirements` path to either a pip- or conda-style requirements file. 

local data still only tracks the most recently selected `--base-image` instead of storing the bespoke local images generated by the new function

## Discussion 
My first few stabs at this I had the CLI either remember the most-recently-used requirements file along with the base image and/or caching the pre-baked image instead, but it felt pretty unwieldy for not a lot of benefit; the layer caching we get for free with docker is basically just as good as caching the custom image in local data. 

The potential downside of this is that users need to specify the `--requirement` argument for both `notebook start` and `notebook publish` every time, which might be surprising because it's different to the `--base-image` behavior. 

## Testing
I tested this manually end-to-end with a dummy entrypoint that just reports the version of a given package in `sys.modules` and sanity-checked with packages installed via a `requirements.txt` and an `environment.yml`. 

I also had chatgpt write me a few happy path unit tests.

## Documentation

No docs updates besides CLI help text / docstrings 

----
📚 Documentation preview 📚: https://garden-ai--373.org.readthedocs.build/en/373/

### DIFF
--- a/garden_ai/app/notebook.py
+++ b/garden_ai/app/notebook.py
@@ -110,19 +110,7 @@ def start(
         custom_image_uri,
         None if need_to_create_notebook else notebook_path,
     )
-
-    # validate requirements file
-    if requirements_path:
-        requirements_path.resolve()
-        if not requirements_path.exists():
-            typer.echo(f"Could not find file: {requirements_path}")
-            raise typer.Exit(1)
-        if requirements_path.suffix not in {".txt", ".yml", ".yaml"}:
-            typer.echo(
-                "Requirements file in unexpected format. "
-                f"Expected one of: .txt, .yml, .yaml; got {requirements_path.name}."
-            )
-            raise typer.Exit(1)
+    _put_notebook_base_image(notebook_path, base_image_uri)
 
     # Now we have all we need to prompt the user to proceed
     if need_to_create_notebook:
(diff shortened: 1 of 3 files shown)


WillEngler:
> The potential downside of this is that users need to specify the --requirement argument for both notebook start and notebook publish every time, which might be surprising because it's different to the --base-image behavior.

Admittedly the --base-image behavior is a little awkward itself. Maybe as part of #362 we should think through both the --base-image and --requirements behavior together. But for now I think your solution is a huge improvement.

> the layer caching we get for free with docker is basically just as good as caching the custom image in local data.

Good realization, and I love that this keeps things pretty simple for our implementation.

WillEngler (approved):
This looks fantastic. It seems like a lot of thought and tinkering went into getting the behavior to feel right from the user's perspective. Great stuff.

PR #393, opened 2024-01-22
383 don't run tests during publication

### LINKED ISSUE #383: Make the @test_function not run during publication
Problem: Users need to test in the notebook to make sure their entrypoint works. But then they have to remember to comment out the function that executes it. This is a common way for people to be tripped up.

Solution: Have the `@test_function` not run during publication time, and make clear to users in the template notebooks that this is the "right" way to do things.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
Closes #383 

## Overview

functions decorated with `@entrypoint_test` now become no-ops at publication (and `notebook debug`) time, so users don't need to remember to comment out tests before they publish. 

## Discussion

`containers.build_notebook_session_image` can now set environment variables in the docker image it builds. This currently just sets a `GARDEN_SKIP_TESTS=True` flag by default, which the `entrypoint_test` decorator now checks for before calling the underlying test function.  

## Testing

Just manual tests so far -- I wasn't sure how useful a unit test would be since this only changes behavior in containers, but I'm happy to add one

As a bonus though, manual testing actually turned up a lil bug in `notebook debug` (we were attempting to pull images by an id which would only exist locally) 🪲 

## Documentation

Updated the template notebooks with a little explanation/usage of the decorator, but no actual docs changes otherwise.

----
📚 Documentation preview 📚: https://garden-ai--393.org.readthedocs.build/en/393/

### DIFF
--- a/garden_ai/app/notebook.py
+++ b/garden_ai/app/notebook.py
@@ -315,7 +315,9 @@ def debug(
             docker_client, base_image, requirements_path
         )
 
-        image = build_notebook_session_image(docker_client, path, local_base_image_id)
+        image = build_notebook_session_image(
+            docker_client, path, local_base_image_id, pull=False
+        )
         if image is None:
             typer.echo("Failed to build image.")
             raise typer.Exit(1)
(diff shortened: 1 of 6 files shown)


WillEngler (approved):
I really like this approach! Good thinking

PR #399, opened 2024-02-05
Make Docker build errors more useful

### LINKED ISSUE #367: Elegantly display build errors during `garden-ai publish`
Error handling in `garden-ai publish` has a feast or famine dynamic. If a fatal error happens while building your container and you haven't used the `--verbose` flag, you basically get no explanation of what went wrong. If you _do_ use the verbose flag, then you get some big typer stack traces and if you scroll up a bit in your terminal you can see the error message you needed to see.

How to repro:
- Use garden-ai start to make a notebook with a glaring Python error. Divide by 0 or something.
- Try to publish it with `garden-ai publish` with and without --verbose

We should give the right error message to explain where a user should try to fix their notebook without `--verbose`. With `--verbose` we should highlight the helpful message and not show irrelevant stack traces.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Closes #367 

## Overview

It was the case that errors during Docker builds were uncaught and surfaced as if they were exceptions thrown by the Garden app. This meant that the user got a big Typer stack trace that had nothing to do with their build issue. 

We expect build errors to happen during normal usage, say if a user has a bug in their notebook or a bad requirements file. This PR makes it so that if there is a build error, we display the full build log so the user can diagnose the issue. We also display a helpful message at the end.

## Discussion

For the case where there is a bug in the user's notebook code, we now put the relevant traceback right where they can see it and act on it.

Screenshot 2024-02-05 at 12 51 09 PM

## Testing

I tested manually with the notebook publish command. I used a notebook where I divided by 0 to test the "build" segment and a known-bad requirements.txt to test the "prebuild" segment. I also added a new unit test.

## Documentation

No docs changes.

----
📚 Documentation preview 📚: https://garden-ai--399.org.readthedocs.build/en/399/


garden_ai/app/notebook.py
```diff
@@ -37,6 +39,57 @@
...
+
+    def handle_docker_build_failure(self, e: docker.errors.BuildError):
+        # If the user is in verbose mode, the build log has already been printed.
+        if not self.verbose:
+            for line in e.build_log:
+                typer.echo(line)
+
+        print_err(f"Fatal Docker build error: {e}\n" "Above is the full build log.\n")
+
+        if type(e) is DockerPreBuildFailure:
```
OwenPriceSkelly, line 71: purely stylistic nit given that `DockerPreBuildFailure/DockerBuildFailure` don't have any subclasses, but should we stick to `isinstance` just for consistency?

garden_ai/app/notebook.py
```diff
@@ -37,6 +39,57 @@
...
+            print_err(
+                "Garden could not set up your base Docker image. "
+                "If you supplied a requirements file, check that it's formatted correctly.\n"
+            )
+        elif type(e) is DockerBuildFailure:
+            last_line = e.build_log[-1] if len(e.build_log) > 0 else None
+            if "Traceback" in last_line:
```
OwenPriceSkelly, line 78: I think this `in last_line` check will break if it really is None

OwenPriceSkelly (approved):
lgtm! handful of nits but nothing worth blocking over

garden_ai/app/notebook.py
WillEngler: Good catch, addressed

garden_ai/app/notebook.py
WillEngler: I'm indifferent, switched to isinstance

garden_ai/containers.py
```diff
@@ -394,22 +400,53 @@ def build_image_with_dependencies(
...
-                raise docker.errors.BuildError(reason=error_message, build_log=stream)
+        image = process_docker_build_stream(
+            stream, client, DockerPreBuildFailure, print_logs
+        )
+        return image.id
+
+
+def process_docker_build_stream(
+    stream: Iterator[dict],
+    client: docker.client,
```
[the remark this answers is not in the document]
WillEngler: Oops, thanks

PR #403, opened 2024-02-06
 `docker prune` command

### LINKED ISSUE #366: `garden-ai purge` to get rid of any local Docker artifacts created/pulled by Garden
Because Garden goes and runs Docker commands on behalf of the user, they will end up with cruft images eventually. There should be a command that removes any Docker images and containers created or pulled by Garden. That way the user doesn't have to fumble around with docker commands to reclaim disk space.

`garden-ai weed` might be fun too 🤔 

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
Closes #366 

## Overview

This implements the `garden-ai docker prune` command to delete any garden-related docker artifacts from the user's local machine. 

## Discussion
We'd been pretty fast and loose with creating containers / images on the user's machine, mostly because docker's layer caching let us get away with it. A lot of the legwork for this one involved testing/fixing mistakes I hadn't really noticed before, like that a failing build would leave intermediate containers behind even if a successful one wouldn't. Besides making the `containers.py` helper functions into better boyscouts, functionality hasn't changed significantly, we just tag everything now.

## Testing

I tested throughout with my local docker install and updated the existing unit tests to cover some of the changes to how we call the docker sdk. 

## Documentation

Documented with the command help text but no other docs 

----
📚 Documentation preview 📚: https://garden-ai--403.org.readthedocs.build/en/403/


garden_ai/app/docker.py
```diff
@@ -48,3 +53,32 @@ def check(
...
+        prefixes += ["gardenai/base"]
+
+    for image in client.images.list():
+        for tag in image.tags:
+            if tag.startswith(tuple(prefixes)):
+                if dry_run:
+                    print(f"Would remove {image.short_id} ({tag})")
+                else:
+                    print(f"Removing {image.short_id} ({tag})")
+                    client.images.remove(tag)
```
WillEngler, line 84: Nice and clean!

garden_ai/app/notebook.py
```diff
@@ -363,7 +363,10 @@ def debug(
         )
 
         image = build_notebook_session_image(
-            docker_client, path, local_base_image_id, pull=False
+            docker_client,
+            path,
+            local_base_image_id,
+            # env_vars={"GARDEN_SKIP_TESTS": False},  # TODO: what's better UX for debugging?
```
WillEngler, line 369: Should this comment be kept?

OwenPriceSkelly, line 369: this was a change I made thinking it might be more convenient, but this could also make the debug image inconsistent with the actual published image

garden_ai/containers.py
```diff
@@ -40,15 +41,19 @@ class DockerBuildFailure(docker.errors.BuildError):
...
     Otherwise the normal Docker exception is still raised.
     """
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         try:
             return func(*args, **kwargs)
+        except docker.errors.BuildError:
+            client = get_docker_client()
+            cleanup_dangling_images(client)
```
WillEngler, line 55: I'm skeptical of how this PR removes _all_ dangling images from the user's machine here and at the start of the prune command. If we knew we were the only application using Docker on their machine I could see it. But I think it's unexpected/surprising behavior that could interfere with other applications.

At most I think we should include a flag in the prune command that enables this. I think we could even get away with not including it at all and pointing users to the existing `docker system prune` command

WillEngler (commented):
I think removing resources that we don't know for sure are Garden resources is dicey, especially when there's an easy way for users to clean it up as is. I don't think we need to take it on. Happy to chat and have my mind changed on this.

garden_ai/containers.py
OwenPriceSkelly: Yeah glad you said something, I was on the fence / don't love the risk of overstepping by cleaning too much. The situation I'm worried about is one where a failed build leaves an untagged image behind which references one of the large base images, then `garden-ai docker prune` cleans out every base image except that one for no obvious reason. 

I could probably filter this exception handling down to "dangling images created since the start of the build process" and skip it in the actual command if that seems sane, or could try and limit it to images that share layers with known garden images

WillEngler (approved):
LGTM!

PR #405, opened 2024-02-09
Let users opt in to when their DOIs are registered

### LINKED ISSUE #400: Opt-in DOI Activation
Garden mints DOIs at creation time for gardens and entrypoints. We want users to be able to explicitly say when they want these DOIs to become "real". (ie, moved out of draft state)

Make two new commands:

```
garden-ai garden make-doi-real
garden-ai entrypoint make-doi-real
```

I don't actually think "make-doi-real" is the right way to convey this. Figure that out as part of the ticket.

Also, update the `garden show/list` and `entrypoint show/list` commands to indicate whether the DOI for each entity has been made official.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Closes #400 

## Overview

This PR adds new commands `garden-ai garden register-doi` and `garden-ai entrypoint register-doi`. These will promote the draft DOIs that Garden mints for all gardens and entrypoints to "registered" and "findable" status in Datacite.

This change is prompted by early feedback that Garden proactively creating and registering DOIs was surprising. This change makes it so that users have the option to review their work before pulling the trigger and making the DOI for an entrypoint or garden official.

The PR also adds to the `list` commands to let users easily see the draft statuses of their DOI.

_Screenshots_

Screenshot 2024-02-08 at 4 18 48 PM

Screenshot 2024-02-08 at 4 41 55 PM

## Discussion

As I was reviewing how we deal with DOIs, it became clear that our support for users bringing their own DOI (not one generated with our account) is broken. This PR does not attempt to fix that. It also doesn't attempt to manage draft status of externally generated DOIs.

## Testing

I tested this manually and added tests for the new commands. I intend to do more end to end testing against the prod DataCite account with [this ticket](https://github.com/Garden-AI/garden/issues/284).

## Documentation

Just inline CLI help for now.

----
📚 Documentation preview 📚: https://garden-ai--405.org.readthedocs.build/en/405/

### DIFF
--- a/garden_ai/app/console.py
+++ b/garden_ai/app/console.py
@@ -6,19 +6,58 @@ from typing import List, Optional, Any
 console = Console()
 
 
+def format_doi_status(doi_status: bool) -> str:
+    return "draft" if doi_status else "registered"
+
+
+# Assuming the boolean value is stored in a field named 'doi_status' in the resource objects
+# Update the _get_rich_resource_table function to include the new column and format the doi_status
 def _get_rich_resource_table(
     resource_objs: Optional[List[Any]], resource_table_cols: List[str], table_name: str
 ) -> Table:
     table = Table(title=table_name)
 
+    # Add all the existing columns
     for col in resource_table_cols:
         table.add_column(col)
+
+    # Add the new column for DOI status
+    table.add_column("DOI Status")
 
     if resource_objs:
         for resource_obj in resource_objs:
             row = []
             for field in resource_table_cols:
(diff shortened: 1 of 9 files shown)


OwenPriceSkelly (approved):
LGTM! should we make a docs or help text note somewhere mentioning that you can still publish/search for/share gardens that only have draft DOIs, you just shouldn't be using it in official citations?

WillEngler:
Yeah, I don't know exactly where and how to put it in the docs yet, but yes we should.

PR #409, opened 2024-02-14
target prod by default

### LINKED ISSUE #284: Make garden-ai point to production by default, including production DataCite account

OwenPriceSkelly:
fixes #284 

## Overview

The default values in `garden_ai.constants` now point at prod instead of dev. This can be controlled with the environment variable `GARDEN_ENV=dev` or `GARDEN_ENV=prod`. 

So that we don't need to worry about accidentally hitting prod ourselves if that env var is left unset, I also added a `.env.shared` file at the project root which sets `GARDEN_ENV=dev`. Because this file isn't distributed to pypi, this only applies if the `garden-ai` executable was installed from source (which is always the case in e.g. a `poetry shell`). 

The `.env.shared` won't override existing values, so this works like you'd expect:  

```sh
$ (plain install) garden-ai do something                # hits prod default 

$ (poetry shell) garden-ai do something                 # hits dev default 
$ (poetry shell) GARDEN_ENV=prod garden-ai do something # hits prod explicitly 
```

For quick sanity checking I also added a bit to `garden-ai --version` so that it'll print `GARDEN_ENV` if it's been set: 

```sh
$ (poetry shell) garden-ai --version
garden-ai 0.0.0 (dev)
```

## Testing

sanity checked DOI minting but haven't done the full end-to-end prod sanity check 

## Documentation

No docs changes, but I could be persuaded to add something to the contributor guide 

----
📚 Documentation preview 📚: https://garden-ai--409.org.readthedocs.build/en/409/

### DIFF
--- a/.env.shared
+++ b/.env.shared
@@ -1,2 +1,3 @@
 # note: does not override existing vars
+# so if you've explicitly set GARDEN_ENV=prod, that choice will be respected
 GARDEN_ENV=dev
--- a/.env.shared
+++ b/.env.shared
@@ -1,0 +1,2 @@
+# note: does not override existing vars
+GARDEN_ENV=dev
--- a/garden_ai/app/main.py
+++ b/garden_ai/app/main.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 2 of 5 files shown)


WillEngler (approved):
Love it, especially how you anticipated the dev experience stuff

Issue #402, opened 2024-02-05
Go do the whole DLHub migration


WillEngler:
We will probably split this ticket as we go and learn more. Make it so that all the DLHub models we want to migrate are available in prod Garden.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Scoping this to the dev dry run. Opened a new ticket to pull the trigger in prod when we're ready.

WillEngler:
Prod dry run completed and migration script merged in with #411

PR #410, opened 2024-02-15
Add logic to format DLHub model input and output correctly

WillEngler:
This came up in the course of #402

## Overview

DLHub wraps user input to models in a dictionary before sending it along to Globus Compute. Similarly, when receiving output from a successful run, DLHub unwraps the output from a nested tuple and returns the inner output to the user.

This PR makes it so that Garden will do the same for the legacy DLHub models we're migrating.

## Discussion

This is a little kludgey but good enough for now. We might be able to do something a little fancier where we add this code to the DLHub functions registered with Globus Compute if/when we want to get rid of this little hack.

## Testing

I tested this against sample input and output I got from a DLHub model that I migrated to Garden dev. I also added new unit tests. We will still need to test this end to end in the context of the full migration dry run on dev.

## Documentation

None

----
📚 Documentation preview 📚: https://garden-ai--410.org.readthedocs.build/en/410/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -635,12 +635,3 @@ class GardenClient:
             if garden:
                 print(f"(Re-)publishing garden {garden.doi} ({garden.title}) ...")
                 self.publish_garden_metadata(garden)
-
-    def make_entrypoint_from_function_id(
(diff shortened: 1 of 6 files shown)


OwenPriceSkelly (approved):
lgtm!

PR #415, opened 2024-02-21
append `:home-run` tag to image uris

### LINKED ISSUE #408: Bake DLHub package installs into DLHub containers
On the existing demo endpoint, we currently run the following command after starting up a container.

```
pip install funcx==1.0.13 funcx-endpoint==1.0.13; pip install home_run==0.5.0; export PYTHONPATH="$PYTHONPATH:/app:/home/ubuntu"; mkdir -p /home/ubuntu/
```

This is needed for DLHub containers. To my knowledge, DLHub did not bake `home_run` or similar into its containers at publish time. So this is necessary for DLHub models to work.

This slows down Garden container start-up time. It also means that DLHub functions that work on the demo endpoint won't work on other K8s endpoints. I think we should run this command as an extra layer on the DLHub containers and host this "batteries included" version on the Garden Docker Hub account.

Related to #407.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
closes #408. part of #402 

## Overview

This just adds `:home-run` tags to the respective image uris, so that the dlhub functions are registered against images with `home_run==0.5.0` installed and we can remove that part from the endpoint-specific worker init. 

----
📚 Documentation preview 📚: https://garden-ai--415.org.readthedocs.build/en/415/

### DIFF
--- a/scripts/dlhub_migrate.py
+++ b/scripts/dlhub_migrate.py
@@ -129,7 +129,7 @@ pacbed_entrypoint_meta = EntrypointMetadata(
 pacbed_bundle = EntrypointBundle(
     pacbed_entrypoint_meta,
     "35090f12-ec2c-4138-8433-2bbdb54768e1",
-    "gardenai/cbc3e248-8b62-4c74-b0ba-ee837197c3be",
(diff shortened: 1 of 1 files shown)


WillEngler (approved):
LGTM!

PR #416, opened 2024-02-22
Extract metadata.json from built container as a file instead of `cat`ting its contents

### LINKED ISSUE #414: The Ryan Jacobs Heisenbug
Our collaborator Ryan from Wisconsin tried using Garden for the first time yesterday. It was going pretty well, right up until the end of the `publish` command. In `extract_metadata_from_image`, Ryan was getting an empty string from the Docker command that `cat`s the metadata.json in the container.

```
container_stdout = client.containers.run(
        image=image.id,
        entrypoint="/bin/sh",
        command=["-c", "cat /garden/metadata.json"],
        remove=True,
        detach=False,
    )
```

His container had already been pushed to our ECR repo. I pulled it down and confirmed that the metadata.json was there. I also confirmed that when _I_ run this code against that container, I get the expected JSON. Owen also could not repro the bug.

So there is something particular to Ryan's computer that makes the exchange over stdout from Docker to the Garden host process not work. Maybe his Docker settings, maybe something OS level, we really don't know. But Owen and I theorized that if we could _not_ rely on stdout, maybe that would solve it. We had Ryan run a [test script](https://gist.github.com/WillEngler/75fcf1d1756887c878d2ac2c364dfbb3) that seems to confirm that theory.

So, the move is probably to use `container.get_archive` to pull files from a container.

# Assumptions:
1. Confirm this doesn't break things for non-Ryans.
2. Cut a release and follow up with Ryan

# Acceptance Criteria
Given I am using Ryan's laptop, when I publish an entrypoint with Garden, then it doesn't break at the metadata parsing step.

WillEngler:
Might close #414 

## Overview

We _think_ that there is a bug on certain computers where stdout is not properly captured in the following line:

```
container_stdout = client.containers.run(
        image=image.id,
        entrypoint="/bin/sh",
        command=["-c", "cat /garden/metadata.json"],
        remove=True,
        detach=False,
    )
```

The Gardeners can't reproduce the bug, but we want to try circumventing stdout altogether to see if that fixes it. To do this we'll extract a file from the container instead of `cat`ting the file contents.

## Testing

I updated the unit test and I did manual runs to confirm it still works

## Documentation

No

----
📚 Documentation preview 📚: https://garden-ai--416.org.readthedocs.build/en/416/

### DIFF
--- a/garden_ai/containers.py
+++ b/garden_ai/containers.py
@@ -300,12 +300,15 @@ def extract_metadata_from_image(
 
         # There is only one file in the tarball
         member = tar.getmembers()[0]
-        file_contents = tar.extractfile(member).read().decode("utf-8")
+        file_obj = tar.extractfile(member)
+        if not file_obj:
+            raise FileNotFoundError(f"Could not find {file_path} in image {image.id}")
+        decoded_file_contents = file_obj.read().decode("utf-8")
     finally:
         container.stop()
         container.remove()
 
-    return json.loads(file_contents)
+    return json.loads(decoded_file_contents)
 
 
 @handle_docker_errors
--- a/garden_ai/containers.py
+++ b/garden_ai/containers.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import os
+import io
 import pathlib
 import tarfile
 import functools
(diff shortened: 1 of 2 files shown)


OwenPriceSkelly (approved):
lgtm! fingers crossed this squashes the heisenbug 🤞

PR #417, opened 2024-02-22
make github connector idempotent

### LINKED ISSUE #413: GitHubConnector.stage() is not idempotent
How to reproduce:
1. Make a GitHubConnector with some GitHub repo.
2. Run .stage() twice.

You will get 
```
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --branch=main -- https://github.com/Garden-AI/garden.git gh_model
  stderr: 'fatal: destination path 'gh_model' already exists and is not an empty directory.
```

This is a problem - we need the .stage operation to be idempotent. We expect .stage() to be called many times in the lifetime of an entrypoint container deployed on an endpoint. It needs to be able to run multiple times.

# Assumptions:
1.

# Acceptance Criteria
Given I am a publisher, when I am writing an entrypoint, then I don't have to worry about checking whether my model files already exist before calling .stage()

OwenPriceSkelly:
fixes #413 

## Discussion

~This tweaks the `trackcalls` decorator so that `stage` can use its own `has_been_called` attribute and guarantee that it only executes once.~

This makes the GH connector behave a little more like the huggingface one by pulling instead of cloning when the directory is already a git repo. 

## Testing

~Wrote a cute little unit test, since it was easy~ 
updated the unit test so it checks that pull is called instead of clone when appropriate
 
## Documentation

no docs 

----
📚 Documentation preview 📚: https://garden-ai--417.org.readthedocs.build/en/417/

### DIFF
--- a/garden_ai/model_connectors/github_conn.py
+++ b/garden_ai/model_connectors/github_conn.py
@@ -50,9 +50,6 @@ class GitHubConnector:
                 repo.remotes.origin.pull(self.branch)
                 return self.local_dir
 
-        if not os.path.exists(self.local_dir):
-            os.mkdir(self.local_dir)
-
         Repo.clone_from(f"{self.repo_url}.git", self.local_dir, branch=self.branch)
 
         if self.enable_imports:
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
This looks great!

PR #429, opened 2024-02-29
`garden-ai garden delete` command

### LINKED ISSUE #421: `garden-ai garden delete` command
It will be nice for users to have the ability to delete a garden. If the garden's DOI is one the user has in their data.json, this will both delete the garden locally and in the search index. If the garden has a registered DOI, the command should refuse to delete the garden. There should also be a hidden `--force` flag or similar that will delete the garden from the search index even if it isn't in the user's data.json. This will be useful for admins.

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a publisher, when I publish a test garden that I don't really want published, then I can `garden-ai garden delete` it.

WillEngler:
Closes #421 

## Overview

This PR adds a new command. `garden-ai garden delete -g 10.23677/some-doi`. It deletes the garden locally and on the search index.

## Discussion

By default, if a garden isn't in your data.json, the command won't let you remove it from the search index. There is a hidden `--dangerous-override` flag you can add that will let you remove it from the search index anyways.

## Testing

I tested this against the new DELETE /garden-search-record endpoint in dev. I also added new unit tests for the command

## Documentation

Just the CLI docs for now

----
📚 Documentation preview 📚: https://garden-ai--429.org.readthedocs.build/en/429/

### DIFF
--- a/garden_ai/app/garden.py
+++ b/garden_ai/app/garden.py
@@ -358,11 +358,10 @@ def delete(
 
         if not garden:
             # It has a registered DOI and they don't have the garden locally - nothing to do
-            typer.confirm(
+            console.print(
                 f"The DOI {garden_id} is registered, so we can't remove it from the search index. "
                 "You also don't have it in your local data. "
                 "There is nothing to delete.",
-                abort=True,
             )
             raise typer.Exit(code=1)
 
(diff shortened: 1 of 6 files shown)


OwenPriceSkelly (approved):
LGTM with just one (1) ask: 

I think attempting to delete a non-draft DOI should fail with an error, at least for now. Especially since those are only going to be cases where the user explicitly opted in to registering the DOI. 

The upshot of this is a) avoid some datacite crimes but also b) nobody else can delete my garden if I've `register-doi`-ed it, poweruser or otherwise.

Longer term, in vRealApp, I think we can relax this when we have real access control and tombstone pages etc

WillEngler:
I like the ask! I'll make that change and then merge

blaiszik:
Came here to mention deletability of Gardens and entry points with registered DOIs, but see you already found that issue! 🙂

PR #432, opened 2024-03-04
Edit entrypoint and garden metadata from the command line

### LINKED ISSUE #386: Users should be able to edit Garden metadata after creating it
How exactly do we want to accomplish this? I don't know. But currently after going through `garden-ai garden create` your only recourse is to open your data.json in a text editor and edit there.

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a publisher who has made a garden, when I notice a typo in the garden's description, then there should be an easy way to edit it.

WillEngler:
Closes #386

## Overview

Adds new `entrypoint edit` and `garden edit` commands.

## Discussion

I'm deliberately making this pretty quick and dirty to avoid overbuilding this before we make an app and make this obsolete. For example, here's the view after you enter the command

Screenshot 2024-03-04 at 11 40 17 AM

You pick a field with an arrow key and enter, then click OK and you can edit the field. Having to click OK is janky, but (surprisingly) removing the need to click OK would be a step up in complexity to implement. This should solve the problem in the short term, even if it isn't beautiful.

## Testing

I tested both commands manually

## Documentation

Just in the CLI help

----
📚 Documentation preview 📚: https://garden-ai--432.org.readthedocs.build/en/432/

### DIFF
--- a/garden_ai/app/entrypoint.py
+++ b/garden_ai/app/entrypoint.py
@@ -5,7 +5,6 @@ import typer
 import rich
 from rich.prompt import Prompt
 
-from garden_ai import local_data
 from garden_ai import GardenClient
 from garden_ai.app.console import (
     console,
@@ -246,6 +245,7 @@ def edit(
 
     entrypoint = get_local_entrypoint_by_doi(doi)
     if not entrypoint:
+        rich.print(f"Could not find entrypoint with doi {doi}")
         raise typer.Exit(code=1)
 
     string_fields = ["title", "description", "year", "short_name"]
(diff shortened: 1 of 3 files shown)


OwenPriceSkelly (approved):
LGTM! a couple gripes but otherwise feels like just the right amount of jank to tide us over to vRealApp
- Can we add a "... then click Ok" phrase to the instructions at the top? I'm not sure I'd have guessed it was clickable otherwise 
- Lets use a non-comma separator (`;`, maybe?) so that people listing authors like `Lastname, Firstname` don't get bit

WillEngler:
Thanks for the feedback. I'll incorporate those suggestions in and then pull this in.

PR #434, opened 2024-03-13
428 reference new endpoint

### LINKED ISSUE #428: update docs/sdk/error handling to reference new demo endpoint
_Before the dlhub endpoint can be switched off, we need to update docs and the sdk in light of the new requirement that users join the correct globus group to use the new demo endpoint._

# Assumptions:
1. the new demo endpoint is functional (#418)
2. demo endpoint is only available to members of the `Garden Users` globus group [here](https://app.globus.org/groups/53952f8a-d592-11ee-9957-193531752178/about) (#425) 
3. no manual approval is needed for users to join the group, just the join link 

# Acceptance Criteria:
- docs/tutorial have a hard-to-miss link to join the group (like MDF does) 
- entrypoints run remotely on demo endpoint by default 
- if the user has not yet joined the group when they try to run an entrypoint, the error message also gives them the link to join

OwenPriceSkelly:
closes #428 

## Overview

This wraps up the demo-endpoint-related changes to the SDK and has the final (for the foreseeable future) tweaks to the helm chart. 

## Discussion

Changes to the workerInit for the demo endpoint have been live since last week -- this is the kludge for running models with python <= 3.7 which I hope to never touch again. 
I also added a 20 minute task timeout, which should be generous enough for good faith demos but not like, mining bitcoin or whatever. 

The docs now point tutorial-followers to join the Garden Users group as both a prerequisite and again at the point where users are going to try running on the demo endpoint.

The special error handling is there so users who bravely don't read docs still have an easy way to join.

still TODO: configure the globus group to auto-accept members instead of (only occasionally?) requiring me to manually approve them, but nothing in this PR is blocked on that. 

## Testing

I tested this with DLHub models (to make sure they remained unbroken) and by logging in with my personal email so I could confirm that the error handling looked correct for group non-members.

## Documentation

Yes, just in the tutorial

----
📚 Documentation preview 📚: https://garden-ai--434.org.readthedocs.build/en/434/

### DIFF
--- a/docs/user_guide/introduction.md
+++ b/docs/user_guide/introduction.md
@@ -1,7 +1,7 @@
 # 🌱 Welcome to Garden 🌱
 
 Docs are WIP, but in the meantime here are good places to start:
-- [Installation](installation.md) - getting your local machine set up for gardening
+- [Installation](installation.md) - getting yourself set up for gardening
 - [Overview](../architecture_overview.md) - High-level explanation of key concepts
 - [Tutorial](../user_guide/tutorial.md) - A short end-to-end example using the CLI. See also: ["seedlings" repository](https://github.com/Garden-AI/seedlings)
 - [contributing](../developer_guide/contributing.md) - Getting started as a contributor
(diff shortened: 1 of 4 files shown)


WillEngler (approved):
LGTM!

Issue #440, opened 2024-03-21
bug installing packages from github in 3.11 notebooks


OwenPriceSkelly:
Noticed this while working on #436. I was sanity checking that the correct version of dill was installed on the endpoint in different python versions, and had modified containers.py so the `RUN pip install garden-ai` would install from my branch instead of pypi. 

while an entrypoint published in a base-3.10 notebook was happy to install garden from my branch, the same entrypoint published in a base-3.11 notebook caused the pod to crash with these logs: 

```
....
Requirement already satisfied: pycparser in /opt/conda/lib/python3.11/site-packages (from cffi>=1.12->cryptography!=3.4.0,>=3.3.1->globus-sdk->globus-compute-endpoint>=2.0) (2.21)
Building wheels for collected packages: pyzmq
  Building wheel for pyzmq (pyproject.toml): started
  Building wheel for pyzmq (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error

  × Building wheel for pyzmq (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [174 lines of output]
      /tmp/pip-build-env-0f6nmyam/overlay/lib/python3.11/site-packages/setuptools/_distutils/dist.py:265: UserWarning: Unknown distribution option: 'cffi_modules'
        warnings.warn(msg)
      running bdist_wheel
      running build
      running build_py
      copying zmq/asyncio.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/_future.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/constants.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/decorators.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/__init__.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/_typing.py -> build/lib.linux-x86_64-cpython-311/zmq
      copying zmq/error.py -> build/lib.linux-x86_64-cpython-311/zmq
      creating build/lib.linux-x86_64-cpython-311/zmq/sugar
      copying zmq/sugar/attrsettr.py -> build/lib.linux-x86_64-cpython-311/zmq/sugar
      copying zmq/sugar/poll.py -> build/lib.linux-x86_64-cpython-311/zmq/sugar
      copying zmq/sugar/context.py -> build/lib.linux-x86_64-cpython-311/zmq/sugar
[...]

Issue #439, opened 2024-03-21
de-conda official base images and bump cryptography requirement


OwenPriceSkelly:
We recently bumped the cryptography library past 42.x, which seems to have a conflict with a conda-managed openssl installed in all of our base images. For now, reverting that change seems like enough of a bandaid, but we should come back to this later. 

If you install garden from github in a notebook and then try to `import garden_ai`, you get something like the following:

```python
File /opt/conda/lib/python3.11/site-packages/OpenSSL/crypto.py:1635, in X509StoreFlags()
   1633 EXPLICIT_POLICY: int = _lib.X509_V_FLAG_EXPLICIT_POLICY
   1634 INHIBIT_MAP: int = _lib.X509_V_FLAG_INHIBIT_MAP
-> 1635 NOTIFY_POLICY: int = _lib.X509_V_FLAG_NOTIFY_POLICY
   1636 CHECK_SS_SIGNATURE: int = _lib.X509_V_FLAG_CHECK_SS_SIGNATURE
   1637 PARTIAL_CHAIN: int = _lib.X509_V_FLAG_PARTIAL_CHAIN

AttributeError: module 'lib' has no attribute 'X509_V_FLAG_NOTIFY_POLICY'
```

but you won't get this if `cryptography<42.0`. You also don't seem to get this if you're in a container without conda.

Hot take: base images shouldn't have a conda install by default. we can offer `:base-*-conda` variety of the base images for the diehards if we really need to, but it makes all of our images heavier than they need to be and is yet another source of mysterious library incompatibilities in the images. 

If we do keep some base images with conda, we'll need to build the images differently than how we currently do so that they're compatible with most recent versions of cryptography etc.

PR #441, opened 2024-03-21
436 relax dill requirement

### LINKED ISSUE #436: maybe relax dill requirement
Ryan Jacobs had a conflict with the version of dill we pin, which was probably more cautious than it needed to be. 

Investigate pinning different known safe versions of dill for the python versions we support instead of just one known safe version. 

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
(hopefully) fixes #436 

## Overview

This relaxes our dill constraints to exactly the same as globus compute. Because we end up installing globus compute _after_ garden anyways (in the endpoint's workerInit), I don't actually think our aggressive pinning strategy was actually _doing_ anything as of upgrading the endpoint. 

I also learned that the version we'd been pinning, dill==0.3.5.1, never supported 3.11 (oops), so I don't think it's even currently possible to publish an image running python 3.11 (dill breaks on the save_session call when I try). 

## Discussion

see #440 -- I'm not sure if this is going to be enough to get images running 3.11 working unless that bug really did have to do with github-related installs. Still, we don't lose any functionality we hadn't already broken by relaxing the constraint here. 

the cryptography de-bump is related to #439 

## Testing

manually, by publishing dummy notebooks with entrypoints that just report the version of dill currently installed. I also made sure I was installing garden from my branch instead of pypi inside of the container. 

----
📚 Documentation preview 📚: https://garden-ai--441.org.readthedocs.build/en/441/

### DIFF
--- a/garden_ai/scripts/save_session_and_metadata.py
+++ b/garden_ai/scripts/save_session_and_metadata.py
@@ -1,11 +1,16 @@
 # the contents of this script are appended to the user's automatically generated
 # notebook script and run in the container in order to persist both a
 # session.pkl and a metadata.json in the final image
-if __name__ == "__main__":
-    # sanity check dill versions
+
+
+def assert_compatible_dill_version():
+    """sanity check that dill version is correct w/r/t the python version.
+
+    This is factored out into its own function because `sys.version_info` does
+    something spooky that breaks dill if it picks up the reference.
+    """
     from importlib.metadata import version
     import sys
-    import dill  # type: ignore
 
     # this should be exactly the same as globus compute requirements
     python_version = sys.version_info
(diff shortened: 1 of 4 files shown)


garden_ai/scripts/save_session_and_metadata.py
```diff
@@ -1,21 +1,42 @@
...
-            "This ensures that the Globus Compute endpoints can deserialize your enviornment. "
-            "Please install dill==0.3.5.1 and try again."
+            "To ensure compatibility with Globus Compute endpoints, Garden needs "
+            f"dill version {required_dill_version} when running python version "
+            f"{'.'.join(map(str, python_version[:2]))}. However, this environment "
+            f"has dill version {dill_version} installed, possibly due to other "
+            "packages installed from within your notebook. \n\nIf you are installing "
+            "dependencies from a cell of your notebook, try specifying them in a "
+            "requirements file with the `--requirements` flag from the command line "
+            "instead try again. "
```
WillEngler, line 29: Love this

WillEngler (approved):
LGTM!

PR #444, opened 2024-03-25
Pick another port if the default Jupyter port is in use

### LINKED ISSUE #395: Gracefully handle the default Jupyter port being in use
Let's say you have a running `notebook start` session. Then in another terminal tab you open another. You will see something like ...

```
APIError: 500 Server Error for
http+docker://localhost/v1.43/containers/6412bc1e0bf85b9377c09e5f4e4337518eb5aa20f8a8b8bae3ee
6180faa626d0/start: Internal Server Error ("driver failed programming external connectivity
on endpoint mystifying_diffie
(0fe999f9f0a389a75574671d76a962d487289200ad21ae3eb8154aca9ac36536): Bind for 0.0.0.0:8888
failed: port is already allocated")
```

This will be another good one to handle. For one, we should pick a higher default port. (Not the standard 8888 that is normally used for Jupyter.) And then at minimum we should print a helpful/informative error when this happens instead of the big stack trace. Even better is if we could dynamically try another port, say `normal_port_number + 1`.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

WillEngler:
Closes #395 

## Overview

This PR makes it so the default port we open Jupyter on is less popular and less likely to have conflicts. (9188 instead of 8888.) And then if 9188 is in use we try 9189, 9190, and so on until we give up at 9198. This makes it so that users can have multiple Jupyter sessions open at a time. It also means that a user running Jupyter on a default port in a standalone application can use Garden without shutting down their other notebook

## Testing

I updated an affected unit test. I also tested manually on cases where I opened multiple instances of Jupyter in multiple terminal windows.

## Documentation

I updated some docs that referenced the old default port.

----
📚 Documentation preview 📚: https://garden-ai--444.org.readthedocs.build/en/444/

### DIFF
--- a/docs/user_guide/faqs.md
+++ b/docs/user_guide/faqs.md
@@ -20,4 +20,4 @@ If calling your entrypoint function in the `garden-ai notebook debug` session be
 
 Alas, you can't. Garden aways spins up Linux containers, so you can't install the `tensorflow-macos` variant. And the normal `tensorflow` doesn't run on Apple silicon.
 
-To work around this, we recommend using a VPS. Spin up a temporary EC2 instance (or equivalent on different cloud providers) running Linux where you can work on your notebook. When you ssh into your remote workstation, forward the port the notebook runs on so that you can still work on the notebook in your local browser. Like `ssh -L 8888:localhost:8888 my-remote-user@my-remote-workstation`.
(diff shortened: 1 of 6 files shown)


OwenPriceSkelly (approved):
LGTM! nice catch updating the port in the docs, too

PR #449, opened 2024-04-02
add missing `manage_projects` scope to auth code

### LINKED ISSUE #448: bug: return of double login for globus compute
> somehow .... ~palpatine~ #209 returned.

Last year, max fixed this with #225, but I've just noticed it again whenever I wipe `~/.garden/tokens.json`. Likely related to bumping all of our packages recently, including `globus-compute-sdk` from 2.11 -> 2.16.

OwenPriceSkelly:
fixes #448 

## Overview
The login bug came back because (as of bumping globus compute from 2.11->2.16) we were no longer requesting a superset of the scopes that the globus compute client does. They now also need the `AuthLoginClient.scopes.manage_projects` scope, so they would trigger another login flow without it. 

Adding the same scope to our list fixed it 🎉

## Discussion

I cleaned up the hardcoded `COMPUTE_RESOURCE_SERVER_NAME = "funcx_service"` to reference the sdk's `ComputeScopes` constant instead, because a comment I found in their sdk indicated they might rebrand that soon too. I also noticed that we're not even requesting our own `action_all` scope (we've just been using the `test_scope`) so I added that to the list, too. 
## Testing

Tested by removing my tokens and logging in again exactly once. Also need to update a unit test. 

## Documentation

no docs

----
📚 Documentation preview 📚: https://garden-ai--449.org.readthedocs.build/en/449/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -11,6 +11,7 @@ from uuid import UUID
 import typer
 from globus_compute_sdk import Client
 from globus_compute_sdk.sdk.login_manager.tokenstore import get_token_storage_adapter
+from globus_compute_sdk.sdk.login_manager import ComputeScopes
 from globus_compute_sdk.serialize.concretes import DillCodeTextInspect
 from globus_sdk import (
     AuthAPIError,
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
LGTM! Good finds

PR #452, opened 2024-04-11
add datasets to entrypoints

### LINKED ISSUE #442: Make it possible for users to add datasets to entrypoints
Currently you can attach datasets to models (and thereby to entrypoints) only by constructing Dataset and Model opjects in the Garden SDK. You can't add a dataset to a model connector, so there isn't an obvious way to add a dataset from a publishing notebook. There also isn't a way to add a dataset to a model or entrypoint via the command line. Assuming our early users will want to add datasets, we'll have to accommodate this somehow.

# Notes:
The fact that datasets are attached to models and not entrypoints makes this awkward from the CLI's perspective. There isn't an obvious way to have a user select a model to attach a dataset to.

# Acceptance Criteria
Given..., when..., then...

OwenPriceSkelly:
closes #442 

## Overview
This PR gives entrypoints a proper `datasets` field (like the existing `papers` and `repositories` fields), which users can populate with the metadata of datasets they want to showcase as related to their entrypoints. 

This keeps the existing `DatasetConnection` object mostly as-is, but I did add a validator to enforce that foundry datasets have both a url and doi. 

Once a user has instantiated `DatasetConnection` object(s) in their notebook, there are a couple ways it can be linked to an entrypoint:
- Directly via the `datasets=[...]` kwarg when instantiating their `EntrypointMetadata` object 
- via a `datasets=[...]` kwarg to the `@garden_entrypoint` decorator 
- (old way) via one of their a model connector's `connector.metadata.datasets` attribute, if they can find it. This is just in here for compatibility.

All three sources of a `DatasetConnection` object will end up in the appropriate `EntrypointMetadata.datasets` attribute (this is handled in the decorator). 

This deliberately does not provide a CLI command to link dataset metadata to an existing entrypoint, because I wanted to avoid the following footgun: publish entrypoint to garden -> add dataset to entrypoint -> forget to re-publish garden from the CLI -> publish another entrypoint to same garden from a notebook -> hey, where'd my dataset metadata go? 

## Discussion 

I'm aiming for the minimum ryan-usable product with this PR, imo linking dataset metadata (or any other kind of related work) is still in "bandaid" territory at this point. I have some half-baked ideas around a more general "`MetadataConnection`" object that we could reuse for papers, datasets, repositories, etc to simplify the way we represent the concept of "related work", but this isn't the PR for that. 

Plan to finish baking that idea when there's a way to interact with / edit metadata from a webapp UI instead of only through the sdk. 

## Testing

I tested this manually by publishing a few entrypoints to the dev search index and verifying that the dataset connection's metadata was included in the entrypoint metadata as expected for datasets linked from either the decorator or constructor. 

I didn't write unit tests for this, bc (a) I felt lazy but also (b) this isn't this something I feel like we need set in stone for the future nor is it really core behavior, so not much benefit in tying it down with unit tests. 

## Documentation

No docs updates along similar reasoning as skipping unit tests, but I could be easily guilted into doing so. otherwise just planning to email ryan with an example so he can keep cookin. 

----
📚 Documentation preview 📚: https://garden-ai--452.org.readthedocs.build/en/452/

### DIFF
--- a/garden_ai/entrypoints.py
+++ b/garden_ai/entrypoints.py
@@ -21,7 +21,7 @@ from garden_ai.datacite import (
     Title,
     Types,
 )
-from garden_ai.mlmodel import ModelMetadata
+from garden_ai.mlmodel import ModelMetadata, DatasetConnection
 from garden_ai.utils.misc import JSON
 
 
@@ -108,8 +108,9 @@ class EntrypointMetadata(BaseModel):
         short_name: This will be the name of the Python method that people call to invoke your entrypoint.
         year: When did you make this entrypoint? (Defaults to current year)
         tags: Helpful tags
-        repositories: List of related code repositories, like GitHub or GitLab repos.
-        papers: List of related papers, like a paper that describes the model you are publishing here.
+        repositories: List of related code repositories (``Repository``), like GitHub or GitLab repos.
+        papers: List of related papers, like a paper (``Paper``) that describes the model you are publishing here.
+        datasets: List of related datasets (``DatasetConnection``) that is related to the entrypoint you are publishing.
     """
 
     doi: Optional[str] = Field(None)
@@ -122,6 +123,7 @@ class EntrypointMetadata(BaseModel):
     models: List[ModelMetadata] = Field(default_factory=list)
     repositories: List[Repository] = Field(default_factory=list)
     papers: List[Paper] = Field(default_factory=list)
+    datasets: List[DatasetConnection] = Field(default_factory=list)
     # The PrivateAttrs below are used internally for publishing.
     _test_functions: List[str] = PrivateAttr(default_factory=list)
     _target_garden_doi: Optional[str] = PrivateAttr(None)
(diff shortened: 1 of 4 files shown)


garden_ai/entrypoints.py
```diff
@@ -275,11 +277,20 @@ def _is_dlhub_entrypoint(self) -> bool:
...
+    datasets: Optional[List[DatasetConnection]] = None,
 ):
     def decorate(func):
+        if datasets:
+            # datasets explicitly connected to entrypoint by decorator
+            metadata.datasets += datasets
         if model_connectors:
-            metadata.models += [connector.metadata for connector in model_connectors]
+            for connector in model_connectors:
+                model_meta: ModelMetadata = connector.metadata
```
WillEngler, line 289: Might not need the model connector case soon

WillEngler (approved):
This seems reasonable! We talked in DMs about consolidating down to One True Way to link datasets so that we don't get bogged down with cruft. But that can come in follow-up PRs.

PR #456, opened 2024-04-19
Create reference docs for garden-ai CLI

### LINKED ISSUE #450: "CLI reference" section in docs
We have a pretty basic "API reference" section in the docs that is generated from docstrings and some of our core object class attributes. (which could also use some polish but that's another issue) 

We don't have anything similar for our CLI, but the CLI commands do (mostly) have good help text which lives in the function docstrings. We should see if there's an easy way to generate a "CLI reference" page from the same docstrings, analogous to the API reference section. 

If there isn't an easy way to do it automatically, we should make a docs page listing every command with at least an example usage for -- users probably won't be aware of certain commands like `garden-ai entrypoint add-paper` or `garden-ai docker prune` otherwise

hholb:
Resolves #450 

## Overview

Create reference documentation for the garden-ai CLI.

## Discussion

Docs were generated using [typer's doc generation utility](https://typer.tiangolo.com/tutorial/package/#generate-docs).

The version of typer we are currently using, 0.7.0, doesn't support this feature. I upgraded to the latest version, 0.12.*, to generate these docs. I ran the test suite using the latest version before reverting back to 0.7.0. One test fails, but it looks like minor change would fix it. Should we consider upgrading to the latest version of typer?

## Testing

Tested using `poetry run mkdocs serve` and it looks good.

## Documentation

This adds a new page for the CLI Reference, and adds an entry to `mkdocs.yml`.

----
📚 Documentation preview 📚: https://garden-ai--456.org.readthedocs.build/en/456/


OwenPriceSkelly (changes requested):
These look great!

Totally fine to change a test that's failing (especially in a docs-only PR where it's clearly not due to any changes to the code's behavior). Does it have to do with the typer version bump?

two very small asks: 
  - can you add a line to `docs/developer_guide/contributing.md` with instructions on how to (re-)build the CLI docs? 
  - it doesn't look like poetry picked up the typer version bump here - you can use `poetry add` (or manually edit the `pyproject.toml`) to reflect the new version of typer we want, then you'll want to run `poetry lock` to rebuild the lockfile (heads up: this [can take a couple minutes](https://python-poetry.org/docs/faq/#why-is-the-dependency-resolution-process-slow)). You can `poetry install` again to keep your environment up to date with the lockfile

OwenPriceSkelly (approved):
LGTM!

PR #458, opened 2024-04-19
@backend_toggle feature flags decorator

### LINKED ISSUE #457: implement feature flags for backend revamp
As we're working on the revamped backend, we want an ergonomic way to control which subset of `BackendClient` methods should target the new backend and which should stay pointed at the current backend.

OwenPriceSkelly:
closes #457 

## Overview

This gives us a nice way to point certain backend client methods at instances of the revamp-in-progress backend for the sake of testing. This is controlled by the values of a new `GARDEN_BACKEND_TOGGLES` environment variable and the old `GARDEN_ENV` variable. 

Note that these are left unset by default if you've e.g. installed garden from PyPI. Default behavior for anyone outside of a `poetry shell` is still always to target the current prod backend. Default behavior for anyone inside a `poetry shell` (or someone who manually sets these variables) has changed.

The new `GARDEN_BACKEND_TOGGLES` variable should be a list of methods to point at revamp resources instead of the usual ones. For example, because I've added a default value of `GARDEN_BACKEND_TOGGLES=mint_doi_on_datacite,update_doi_on_datacite` to the `.env.shared`, if you're running garden from `poetry shell` it will default to the dev lambda (as usual) for every method except the two doi methods, which will now target the dev lightsail deployment.

The existing `GARDEN_ENV` variable can now be `"local"` in addition to "dev" and "prod". If set to dev or prod, the toggled methods will target the respective lightsail deployments instead of the respective dev/prod lambdas. Note that only explicitly setting GARDEN_ENV=prod has this behavior; if GARDEN_ENV is unset then everything is still the old prod resources. 

If set to "local", toggled methods will target `localhost:5500` to play nicely with the revamp's `run-dev-server.sh` script. un-toggled methods will target the dev lambda when GARDEN_ENV=local.

## Discussion

The decorator was probably more fun to write than to read. But it works pretty well and will be nice and easy to delete when we're done with the revamp.

## Testing

Manually, both by exercising backend methods directly from a repl and by using the CLI in a poetry shell.

## Documentation

No docs

----
📚 Documentation preview 📚: https://garden-ai--458.org.readthedocs.build/en/458/

### DIFF
--- a/.env.shared
+++ b/.env.shared
@@ -1,3 +1,4 @@
 # note: does not override existing vars
 # so if you've explicitly set GARDEN_ENV=prod, that choice will be respected
 GARDEN_ENV=dev
+GARDEN_BACKEND_TOGGLES=mint_doi_on_datacite,update_doi_on_datacite
--- a/garden_ai/backend_client.py
+++ b/garden_ai/backend_client.py
@@ -40,24 +40,24 @@ def backend_toggle(func):
     if garden_env not in env_url_mapping or func.__name__ not in toggles:
         return func
 
-    old_url = GardenConstants.GARDEN_ENDPOINT
-    new_url = env_url_mapping[garden_env]
-
-    old_call_method = BackendClient._call
-
-    def _call_with_trailing_slash(self: BackendClient, http_verb, resource, payload):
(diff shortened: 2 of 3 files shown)


WillEngler (approved):
LGTM

PR #467, opened 2024-05-20
Add idempotency check to @entrypoint_test

### LINKED ISSUE #424: Make @test_function run twice to enforce idempotency of endpoints
Our first outside user hit an issue where he got his entrypoint working and published it, but then it was breaking when it was deployed. The entrypoint looked like this ...

```
@garden_entrypoint(
    metadata=entrypoint_meta, 
    model_connectors=[github_asr_connector],
    garden_doi=my_garden_doi
)
def predict_asr(comp_list, elec_list):
    import shutil, zipfile, sys, os
    download_path = github_asr_connector.stage()
    shutil.move('gh_model', 'ASR_model')
    with zipfile.ZipFile('ASR_model/Stability_model.zip', 'r') as zip_ref:
        zip_ref.extractall()

    shutil.move('Stability_model', 'ASR_model') 
    sys.path.append('ASR_model')
    os.system('pip install -r ASR_model/requirements.txt')   
    from model_predict_df import make_predictions
    preds = make_predictions(comp_list, elec_list) 
    return preds
```

On the first run it works fine. On subsequent runs it fails. `zipfile.ZipFile('ASR_model/Stability_model.zip', 'r')` complains that the directory it's trying to extract to already exists.

This shows that it's easy for users to shoot themselves in the foot with file operations or other non-idempotent operations in their entrypoint function. To mitigate this, we can hack the `@test_function` helper so that it runs an entrypoint twice. This will make sure that users see this problem in their code before publishing.

The fact that Garden does not wipe away the container environment between calls is also worth making clear in the docs somewhere.

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

hholb:
Resolves #424

## Overview

`@entrypoint_test` will raise a `EntrypointIdempotencyError` if the function being tested is non-idempotent. If any other exceptions are raised during the test, they are raised as an `EntrypointTestError`. This should make it easier for users to catch issues with their entrypoints before publishing.

## Testing

Add unit tests in `test/test_entrypoints.py` to confirm behavior.

## Documentation

Updates the doc comments in `entrypoint_test`

### DIFF
--- a/garden_ai/entrypoints.py
+++ b/garden_ai/entrypoints.py
@@ -274,6 +274,18 @@ class RegisteredEntrypoint(EntrypointMetadata):
         return self.doi in GardenConstants.DLHUB_DOIS
 
 
+class EntrypointIdempotencyError(Exception):
+    """Raised when an entrypoint function is found to be non-idempotent."""
+
+    pass
+
+
+class EntrypointTestError(Exception):
+    """Raised when an error occurs testing an entrypoint function."""
+
+    pass
+
+
 def garden_entrypoint(
     metadata: EntrypointMetadata,
     garden_doi: str = None,
@@ -325,6 +337,10 @@ def entrypoint_test(entrypoint_func: Callable):
             return results
 
         ```
+
+    Raises:
+        EntrypointTestError: When an error occurs testing the entrypoint_func.
+        EntrypointIdempotencyError: When entrypoint_func is found to be non-idempotent.
     """
     if not entrypoint_func or not entrypoint_func._garden_entrypoint:  # type: ignore
         raise ValueError("Please pass in a valid entrypoint function")
@@ -342,7 +358,17 @@ def entrypoint_test(entrypoint_func: Callable):
             if os.environ.get("GARDEN_SKIP_TESTS") == str(True):
                 return None
             else:
-                return test_func(*args, **kwargs)
+                # call the test_func once
+                result = test_func(*args, **kwargs)
+                try:
+                    # Call the test_func again with the same args to enforce idempotency.
+                    if result != test_func(*args, **kwargs):
+                        raise EntrypointIdempotencyError(
+                            "Your entrypoint function is not indempotent."
+                        )
+                    return result
+                except Exception as e:
+                    raise EntrypointTestError() from e
 
         return inner
 
--- a/tests/test_entrypoints.py
+++ b/tests/test_entrypoints.py
@@ -1,6 +1,10 @@
+import unittest
 from unittest.mock import patch, MagicMock
+from garden_ai import EntrypointMetadata, garden_entrypoint, entrypoint_test
 from garden_ai.entrypoints import (
     RegisteredEntrypoint,
+    EntrypointIdempotencyError,
+    EntrypointTestError,
 )  # Adjust import paths as necessary
 
 # Mock UUIDs for testing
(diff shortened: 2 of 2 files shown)


garden_ai/entrypoints.py
```diff
@@ -274,6 +274,18 @@ def _is_dlhub_entrypoint(self) -> bool:
         return self.doi in GardenConstants.DLHUB_DOIS
 
 
+class EntrypointIdempotencyError(Exception):
+    """Raised when an entrypoint function is found to be non-idempotent."""
+
+    pass
+
+
+class EntrypointTestError(Exception):
```
OwenPriceSkelly, line 283: The `EntrypointIdempotencyError` exception is great but I don't think this one is pulling its weight.

OwenPriceSkelly: Not to belabor the point but in case it's instructive, here's how I'm thinking about it. The main benefits we get from having fine-grained exceptions are:
  - better structure/information in the stacktraces -- it's easier to quickly tell where to start debugging when the bottom of the stacktrace says "EntrypointIdempotencyError" instead of a generic "ValueError" where you have to read up through a few frames of context to see what's happening
  - if we build code on top of this that can tolerate the exception, we want to handle _just_ the specific exception, and let any broader exceptions which we didn't specifically handle propagate up naturally. 
  
So I think this "doesn't pull its weight" in the sense that (a) a user confronted with an `EntrypointTestError` doesn't get more info out of the stacktrace than they would've from the original exception (especially here, since the original came from the user's code) and (b) because it's so general, we can't write more granular exception handling code around it than we could have with a bare Exception

OwenPriceSkelly: for the record I think that the second point doesn't really apply here for either `EntrypointIdempotencyError` or `EntrypointTestError` because they're raised around arbitrary user code, but `EntrypointIdempotencyError` _would_ get us better stacktraces so it's Good

garden_ai/entrypoints.py
```diff
@@ -342,7 +358,17 @@ def inner(*args, **kwargs):
...
+                # call the test_func once
+                result = test_func(*args, **kwargs)
+                try:
+                    # Call the test_func again with the same args to enforce idempotency.
+                    if result != test_func(*args, **kwargs):
+                        raise EntrypointIdempotencyError(
+                            "Your entrypoint function is not indempotent."
+                        )
+                    return result
+                except Exception as e:
```
OwenPriceSkelly, line 370: I don't hate the pattern of raising an error from the try block just to except it immediately, but raising `EntrypointTestError() from e` when `e` could already be an `EntrypointIdempotencyError` is an unnecessary middleman imo

OwenPriceSkelly (changes requested):
This is a good first pass -- love the tests, as always -- but I think we can simplify the errors (more context below)

garden_ai/entrypoints.py
hholb: That makes sense. Thank you for the feedback!

garden_ai/entrypoints.py
hholb: I see what you are saying, and yeah, it doesn't make much sense to raise an error just to immediately catch it and throw another. My intention was to catch other errors that might occur when calling the entrypoint a second time, but I think it will be more informative to just let those propagate on their own.

OwenPriceSkelly (approved):
LGTM

PR #470, opened 2024-05-23
Add requirements to RegisteredEntrypoint metadata

### LINKED ISSUE #375: Include requirements files in metadata
As of #336, users can supply requirements files to the `notebook` commands. We should save this as metadata for entrypoints just as we currently do for the Docker image locations and such.

# Assumptions:
1. The requirements files should be small enough that we can include them in the search payloads. Maybe impose a sanity check size limit so users can't accidentally break their gardens.
2.

# Acceptance Criteria
Given I am interested in learning more about someone's entrypoint that used a `requirements` file, when I pull down a search entry for that entrypoint, then the requirements file they specified is included.

hholb:
Resolves #375
 
## Overview

- Pin requirements in the `RegisteredEntrypoint` metadata if a requirements file is present when a notebook is published.

- Write the requirements into `metadata.json` inside the published container image. This ensures when a garden is published to the search index the metadata for each entrypoint includes the requirements.

- Fix a small bug where creating a garden with no contributors threw an exception.

- add flake8 to pre-commit

- add emacs TAGS to `.gitignore`

## Testing

Tested manually by creating a notebook with a requirements file and a single entrypoint, publishing it, and confirming that the `metadata.json` in the published image contains the requirements data. Tested another notebook with no added requirements to ensure other functionality is not impacted.

I am struggling to come up with efficient unit tests for this. I want to confirm the `metadata.json` is correct, but to do that I need to build a container image which will significantly slow down the test suite.

## Documentation

Updates the doc comments for `RegisteredEntrypoint`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,6 @@ dmypy.json
 
 # VSCode config
 .vscode/*
+
+# Emacs
+TAGS
--- a/garden_ai/app/garden.py
+++ b/garden_ai/app/garden.py
@@ -130,6 +130,8 @@ def create(
                     authors += [name]
                 else:
                     break
+        else:
+            contributors = []
 
     if not description:
         description = Prompt.ask(
--- a/garden_ai/scripts/save_session_and_metadata.py
+++ b/garden_ai/scripts/save_session_and_metadata.py
@@ -31,6 +31,37 @@ def assert_compatible_dill_version():
         raise EnvironmentError(message)
 
 
+def get_requirements_file():
+    """Return pathlib.Path to a requirements file in the container if one exists.
+
+    Requirements are written into the container in the same directory as the notebook.
+    See garden_ai/app/notebook.py `publish`.
+    """
+    from pathlib import Path
+
+    pip = Path(__file__).parent / "requirements.txt"
+    if pip.exists():
+        return pip
+
+    conda = Path(__file__).parent / "requirements.yml"
+    if conda.exists():
+        return conda
+
+    return None
+
+
+def get_requirements_data():
+    """Return a list of requirements for the current notebook."""
+    from garden_ai.notebook_metadata import read_requirements_data
+
+    path = get_requirements_file()
+    if not path:
+        return None
+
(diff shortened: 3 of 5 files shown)


MaxTuecke:
LGTM!

MaxTuecke (approved):
LGTM!

.gitignore
```diff
@@ -140,3 +140,6 @@ dmypy.json
 
 # VSCode config
 .vscode/*
+
+# Emacs
```
OwenPriceSkelly, line 144: 😤

garden_ai/scripts/save_session_and_metadata.py
```diff
@@ -31,6 +31,37 @@ def assert_compatible_dill_version():
...
+    Requirements are written into the container in the same directory as the notebook.
+    See garden_ai/app/notebook.py `publish`.
+    """
+    from pathlib import Path
+
+    pip = Path(__file__).parent / "requirements.txt"
+    if pip.exists():
+        return pip
+
+    conda = Path(__file__).parent / "requirements.yml"
```
OwenPriceSkelly, line 46: easy to be misled because I think we still have some conda-related cruft laying around the repo, but we actually don't currently support conda in the base images so we only need to worry about pip

garden_ai/app/garden.py
```diff
@@ -130,6 +130,8 @@ def create(
                     authors += [name]
                 else:
                     break
+        else:
+            contributors = []
```
OwenPriceSkelly, line 134: good catch. surprised I haven't been burned by this yet

OwenPriceSkelly (approved):
LGTM! 

w/r/t the difficulty with unit testing this -- totally fair, especially when it ended up being a small change to some already hard-to-test code. I think you made the right tradeoff, unit tests should be something you can run fast; in an ideal world this is the sort of behavior that we'd cover in an integration test suite that we could run at regular intervals but outside of the unit tests

p.s. congrats on successfully braving save_session_and_metadata, definitely the weirdest part of the repo imo

PR #476, opened 2024-06-04
Bump Pydantic to 2.7.2

### LINKED ISSUE #228: bump pydantic to 2.x
pydantic v2 is out - among other things, the support for computed properties w validation looks nice, probably some other stuff too.  

migration guide: https://docs.pydantic.dev/latest/migration/

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

MaxTuecke:
`Fixes #228`

## Overview

Bumps pydantic from 1.10.13 to 2.7.2
Bumps mypy from 0.981 to 1.10.0

## Discussion

- Added helper `unique_items_validator` in utils/pydantic.py to replace the deprecated `unique_items` argument for pydantic fields
- Added helper `const_item_validator` in utils/pydantic.py to replace the deprecated `const` argument for pydantic fields
- Changed to strict Optional typing since mypy 1.10 has changed its default to `no_implicit_optional=True`.

## Testing
All unit tests passing with no new warnings.
Old gardens / entrypoints are still working with the new schemas.


garden_ai/utils/pydantic.py
```diff
@@ -0,0 +1,27 @@
...
+T = TypeVar("T")
+
+
+def unique_items_validator(v: Optional[List[T]]) -> Optional[List[T]]:
+    if v is None:
+        return v
+    try:
+        if len(v) != len(set(v)):
+            raise PydanticCustomError("unique_list", "list must be unique")
+        return v
```
hholb, line 14: I like this, clever use of a set!

hholb (approved):
LGTM! :rocket:

garden_ai/gardens.py
```diff
@@ -93,30 +104,38 @@ class Garden(BaseModel):
         build one for the user with the datacite api.
     """
 
+    model_config = ConfigDict(validate_default=False)
```
OwenPriceSkelly, line 107: What's the difference between 
`class MyModel(BaseModel, **kwargs)` and `model_config = ConfigDict(**kwargs)`? if they're doing the same thing lets stick to one or the other. either one is fine but aesthetically I do think the inheritance kwargs are Cool

OwenPriceSkelly (approved):
LGTM! very happy to finally have this finally in, and it looks nice and surgical.

garden_ai/gardens.py
MaxTuecke: The docs show example usage of `validate_assignment` as an inheritance kwarg, but overall, it looks like the general recommend practice is using the ConfigDict. So I think we just stick to using just that.

PR #477, opened 2024-06-05
Bugfix: Send correct URL to Datacite on `garden register-doi`

### LINKED ISSUE #474: bug: doi.org redirect points to invalid thegardens.ai url
When a user officially registers a doi with `garden-ai garden register-doi 10.1234/my-fake-doi` (so that it is no longer a "draft doi"), datacite will redirect https://doi.org/10.1234/my-fake-doi to a url we provide. But the url we give datacite is just a naive https://thegardens.ai/10.1234/my-fake-doi, which is incorrect. 

For example, Ryan Jacob's materials properties garden (10.26311/ep98-br79) was registered with datacite, so https://doi.org/10.26311/ep98-br79 redirects to https://thegardens.ai/10.26311/ep98-br79, which is a 404 because the garden _actually_ lives at https://thegardens.ai/#/garden/10.26311%2fep98-br79 

Acceptance Criteria: 
 - clicking https://doi.org/10.26311/ep98-br79 redirects you to the correct thegardens.ai page

hholb:
Fixes #474

## Overview

Send datacite the correctly formatted URL when registering a doi.

## Testing

Added a new unit test in `test/test_client.py` to verify the URL is correct.

----
📚 Documentation preview 📚: https://garden-ai--477.org.readthedocs.build/en/477/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -311,7 +311,11 @@ class GardenClient:
         # "publish" in the event field moves the DOI from draft state to findable state
         # https://support.datacite.org/docs/how-do-i-make-a-findable-doi-with-the-rest-api
         if register_doi:
-            metadata.update(event="publish", url=f"https://thegardens.ai/{obj.doi}")
+            doi_split = obj.doi.split("/")
+            metadata.update(
+                event="publish",
+                url=f"https://thegardens.ai/#/garden/{doi_split[0]}%2f{doi_split[1]}",
+            )
 
         payload = {"data": {"type": "dois", "attributes": metadata}}
         self.backend_client.update_doi_on_datacite(payload)
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -7,6 +7,7 @@ import base64
 from pathlib import Path
 from typing import List, Union
 from uuid import UUID
+import urllib
 
 import typer
 from globus_compute_sdk import Client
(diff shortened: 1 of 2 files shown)


MaxTuecke (approved):
LGTM! 🚀

garden_ai/client.py
```diff
@@ -311,7 +311,11 @@ def _update_datacite(
         # "publish" in the event field moves the DOI from draft state to findable state
         # https://support.datacite.org/docs/how-do-i-make-a-findable-doi-with-the-rest-api
         if register_doi:
-            metadata.update(event="publish", url=f"https://thegardens.ai/{obj.doi}")
+            doi_split = obj.doi.split("/")
+            metadata.update(
```
OwenPriceSkelly, line 315: This is kind of nitpicky because I'm fairly certain that 100% of our DOIs will be minted following the 10.1234/abc-def convention, but I think the spec technically allows for arbitrarily many `/` in the suffix. The builtin `urllib.parse.quote` is a safer bet here to encode the doi string (and is generally very handy / good to know about)

tests/test_client.py
```diff
@@ -198,3 +199,70 @@ def test_client_credentials_grant(cc_grant_tuple):
     assert gc.auth_client.oauth2_validate_token(gc.garden_authorizer.access_token)[
         "active"
     ]
+
+
+def test_client_datacite_url_correct(
```
OwenPriceSkelly, line 204: dang, this is a beast of a unit test -- is this extra auth mocking because we couldn't instantiate the GardenClient without it?

OwenPriceSkelly: ok I thought this smelled familiar -- we do actually have an old `garden_client` fixture in `tests/conftest.py` that does a lot of similar setup -- if that fixture doesn't "just work" here (v possible, I think it predates some of the newer auth code in the client) let's update that fixture and ~unburden this test

OwenPriceSkelly (changes requested):
This is a totally sane first pass, thanks for jumping on this! I've still got a couple pointers but they shouldn't take long -- if my unit test / fixture suggestion doesn't work (or breaks other tests etc) don't feel like you need to sink a ton more time into it.

tests/test_client.py
hholb: Yeah I needed it to construct the client. I was able to run it without all of the auth mocks locally, but it failed running via github actions.  I'll check out the fixture!

OwenPriceSkelly (approved):
LGTM!

PR #478, opened 2024-06-05
Hholb/422 connector design

### LINKED ISSUE #422: A second pass at ModelConnector design
Back when I first made the Hugging Face ModelConnector, it was meant as a quick experiment to see if that interface would work out. It has stuck! Now we should take a second pass and think through things like ...
- What should the directory structure look like when there's more than one ModelConnector being used?
- Should we have a base class that enforces the right interface?

etc

# Assumptions:
1.
2.

# Acceptance Criteria
Given..., when..., then...

hholb:
Resolves #422

## Overview
:exclamation: BREAKING CHANGES :exclamation: 

Redesign existing model connectors to inherit from a new `ModelConnector` abstract base class. The new class leverages pydantic v2 for validation and parsing of user input. This introduces a breaking change because the constructors for the `GitHubConnector` and `HFConnector` no longer accept positional arguments and require keyword args instead.

This no longer works and will raise an error:
```python
c = HFConnector("Garden-AI/sklearn-iris")
```
Now it must be:
```python
c = HFConnector(repo_id="Garden-AI/sklearn-iris")
```

## Discussion

I created a new interface for users to construct these classes: a function called `create_connector` that takes a URL to a repo and returns the appropriate connector type. It is defined in `garden_ai/model_connectors/model_utils.py`

Instead of constructing the correct connector class directly users can now do this:
```python
from garden_ai.model_connectors import create_connector

c = create_connector("https://huggingface.co/Garden-AI/sklearn-iris")
# or
c = create_connector("https://github.com/Garden-AI/garden")
```

I suggest we encourage users to use `create_connector` going forward as that will allow us to change the model connectors more easily without impacting user code. I updated the notebook templates to use `create_connector` as part of this PR.

## Testing

Added new unit tests.
Updated existing unit tests to handle the new constructors.

Can be tested pretty easily in Ipython by importing `create_connector` or the concrete model connectors and giving them a test drive.

## Documentation

Added doc comments for all new functions and classes.

----
📚 Documentation preview 📚: https://garden-ai--478.org.readthedocs.build/en/478/


garden_ai/notebook_templates/sklearn.ipynb
```diff
@@ -47,23 +58,55 @@
...
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {
     "id": "b0s7Bealdp8M"
    },
-   "outputs": [],
+   "outputs": [
```
MaxTuecke, line 77: I think we probably don't want any execution outputs in our template notebooks.

tests/test_models.py
```diff
@@ -261,6 +272,62 @@ def test_GitHubConnector_raises_exception_with_lfs_file(mocker):
...
+    gh = create_connector("https://github.com/real/repo", revision=fake_revision)
+    assert isinstance(gh, GitHubConnector)
+    assert gh.repo_id == "real/repo"
+
+    # URL must be full HTTP e.g. "https://github.com/owner/repo"
+    with unittest.TestCase().assertRaises(UnsupportedConnectorError):
+        create_connector("github.com/bad/repo")
+
+    with unittest.TestCase().assertRaises(UnsupportedConnectorError):
+        create_connector("huggingface.co/bad/repo")
```
MaxTuecke, line 333: Good unit tests!

pyproject.toml
```diff
@@ -90,6 +91,7 @@ mkdocs-extensions = "^0.1.2"
 pymdown-extensions = "^10.0.1"
 mkdocs-material = "^9.1.13"
 mkdocstrings = {extras = ["python"], version = "^0.21.2"}
+jupyter = "^1.0.0"
```
MaxTuecke, line 94: Im getting a very odd error with docker. If I run `garden-ai docker check` it throws the error:
Garden can't access Docker on your computer.
This doesn't look like one of the typical error cases. Printing error from Docker:
Error while fetching server API version: Not supported URL scheme http+docker

As suggested [here](https://github.com/geerlingguy/ansible-role-docker/issues/462), I downgraded from requests==2.32.3 to requests==2.31.0 and that seemed to fix it for me. Id say pin requests to 2.31.0 for now just to be safe.

garden_ai/notebook_templates/sklearn.ipynb
```diff
@@ -85,13 +128,30 @@
...
+       "README.md:   0%|          | 0.00/755 [00:00<?, ?B/s]"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
    "source": [
-    "my_hugging_face_repo = HFConnector(\"garden-ai/sklearn-iris\")"
+    "my_hugging_face_repo = create_connector(\"garden-ai/sklearn-iris\", \"HF\")\n",
```
MaxTuecke, line 152: I think this needs to be `my_hugging_face_repo = create_connector("garden-ai/sklearn-iris", repo_type="HF")`

MaxTuecke (changes requested):
Overall, very good. No qualms with any of the design choices / implementation. See above comments for suggested changes and then it should be good to go.

garden_ai/notebook_templates/sklearn.ipynb
hholb: Thanks for catching this! The repo type was positional at first and I forgot to look back at the notebooks after I changed it to a kwarg.

tests/test_models.py
hholb: Thanks!

MaxTuecke (approved):

garden_ai/model_connectors/model_connector.py
```diff
@@ -0,0 +1,366 @@
...
+    metadata: Optional[ModelMetadata] = None
+    readme: Optional[str] = None
+    model_dir: Optional[Union[Path, str]] = "models"
+
+    model_config = ConfigDict(
+        validate_assignment=True,
+        protected_namespaces=(),
+    )
+
+    def __init__(self, **kwargs):
```
OwenPriceSkelly, line 171: Hot take: I think that we might want to express this init logic as a [pydantic model_validator](https://docs.pydantic.dev/latest/concepts/validators/#model-validators). It's not a super common pattern but I think it's warranted here for the following reason: most of these `Optional` attributes aren't really optional at all, because your code here is guaranteeing that they're not None once they've been initialized. 

 For example, if we put your logic around populating `repo_url` from `repo_id` or vice versa into a "before" validator, we could safely assume that both of those attributes exist and are well-formed everywhere else (i.e. in any field-specific validators).

garden_ai/model_connectors/model_connector.py
```diff
@@ -0,0 +1,366 @@
...
+        """Return a full URL to the repo from self.repo_id."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _checkout_revision(self):
+        """Checkout self.revision if appropriate."""
+        raise NotImplementedError()
+
+    @staticmethod
+    def parse_id_from_url(url: str) -> Optional[str]:
```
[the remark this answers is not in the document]
OwenPriceSkelly: (wrote these before my comment about using a `model_validator` for the init logic -- probably worth syncing up to make sure I'm thinking through this all the way / not sending you down the wrong path here)

garden_ai/model_connectors/model_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+# Maps ModelRepositorys to their corresponding connector class.
+CONNECTOR_MAPPING: Dict[ModelRepository, Type[ModelConnector]] = {
+    ModelRepository.GITHUB: GitHubConnector,
+    ModelRepository.HUGGING_FACE: HFConnector,
+}
+
+
+def match_connector_type_by_url(url: str) -> Optional[Type[ModelConnector]]:
```
OwenPriceSkelly, line 23: since we probably don't need them to be used outside of this module, I like the convention of prefixing helpers with an underscore to signify that they're internal

garden_ai/model_connectors/model_utils.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                raise UnsupportedConnectorError(
+                    "Unable to create ModelConnector.\n"
+                    "Excpected ether URL or repo_id and repo_type.\n"
+                    "Try:\n"
+                    "\tc = create_connector('https:github.com/owner/repo')\n"
+                    "or:\n"
+                    "\tc = create_connector('owner/repo', 'GH')"
+                )
+        except Exception as e:
```
OwenPriceSkelly, line 116: I don't think we need the nested try-block here if we're just re-raising the exception -- it's the same as just letting an exception occur from inside the `except ConnectorInvalidUrlError:` block here

garden_ai/model_connectors/model_utils.py
```diff
@@ -0,0 +1,117 @@
...
+            return CONNECTOR_MAPPING.get(ModelRepository.GITHUB)
+        if repo_type == "HF":
+            return CONNECTOR_MAPPING.get(ModelRepository.HUGGING_FACE)
+        else:
+            raise ValueError("Unsupported repo_type")
+    else:
+        raise ValueError("Invalid repo_id.")
+
+
+def create_connector(
```
OwenPriceSkelly, line 71: I've mentioned it already but I'm a big fan of this UX. I feel like the `create_connector(my_url)` usage is going to be how 99% of model connectors are created now

OwenPriceSkelly: another hot take: I like the url-only instantiation so much I almost think we could unburden this by not supporting `repo_type='HF'` mode of calling this. Do we need to support `create_connector("Garden-AI/sklearn-iris", repo_type='HF')` when users that want more control and/or don't want to instantiate from a URL can directly instantiate with `HuggingFaceConnector("Garden-AI/sklearn-iris")`?

tests/test_models.py
OwenPriceSkelly, line 1: as usual, love the tests!

OwenPriceSkelly (approved):
LGTM! merge away 🫡

Issue #475, opened 2024-05-31
enhancement: notebook metadata editor widget


OwenPriceSkelly:
Our current way of saving metadata in notebooks is a bit clunky. It would be nice if we had a widget that a user could easily edit their notebook's metadata with.

PR #482, opened 2024-06-11
Notebook metadata editor widget

MaxTuecke:
`Adds #475`

## Overview

Adds `display_metadata_widget()`, which displays an ipywidget form that lets users easily edit a notebook's metadata.

## Discussion

In place of the old top cell containing the commented out metadata json, the new top cell of a notebook will import and run `display_metadata_widget()`.

If the user makes any changes to their notebooks requirements, the widget will show a 'Install new requirements' button that installs the new requirements to the container and updates your local requirements file if one was provided.

When a user updates their notebook's metadata with the widget, it will immediately json save the new `NotebookMetadata` object to `/garden/notebook_metadata.json`

In order to then save the metadata to the notebook, jupyter is now being launched with a custom config that adds a post save hook. When the user saves their notebook, the hook goes and looks for the json `NotebookMetadata` file and saves the garden metadata with the rest of the notebook's metadata. Jupyter can get upset about file versions if we just used `nbformat` to add the metadata, so the post save hook is the best way to edit the notebook's file while the notebook is running.

Other changes:
- `build_notebook_session_image` now copies in `custom_jupyter_config.py` to the container so jupyter can be launched with the custom config. The last line of this file `c.FileContentsManager.post_save_hook = post_save_hook` is appended to the file during runtime, since jupyter's magic variable `c` causes an error when trying to import the script.
- `start_container_with_notebook` has a new argument, `custom_config`, which is default True. This adds 
`--config=/garden/custom_jupyter_config.py` to the entrypoint command for the container. The only time this would be turned off is when using `notebook debug`. Otherwise, the post save hooks should always be turned on.
- `start_container_with_notebook` now launches the container with a new env vars `NOTEBOOK_PATH` and `REQUIREMENTS_PATH` which is the notebook's and requirements (if one was provided) path inside the container. It's a pain to get the either while it is running, so this just makes life easier.
- `start_container_with_notebook` is binding the users requirements file (if one was provided) to the one in the container.
- The 'sister' python script that is made during `build_notebook_session_image` will exclude the tagged widget cell from the script.
- Changed `NotebookMetadata` and `RequirementsData` from namedtuple to pydantic BaseModel.
- `notebook start` will now always try to add the widget cell and empty garden_metadata dict to the notebook if either is missing. The widget cell is tagged with `garden_display_metadata_cell`
- Updated the notebook templates with the new format.

## Testing

Updated the old test_metadata tests to test for the new format.

----
📚 Documentation preview 📚: https://garden-ai--482.org.readthedocs.build/en/482/


garden_ai/notebook_metadata.py
```diff
@@ -221,6 +177,126 @@ def save_requirements_data(
...
 
 
+def display_metadata_widget():
+    """
+    Displays the metadata editor widget
+    When one of the widgets fields is changed, pickles and saves the updated NotebookMetadata.
+    When the notebook is saved, the post_save_hook in custom_jupyter_config will
+    go and look for the pickled NotebookMetadata and save it to the notebooks metadata.
+    """
+
```
hholb, line 187: This looks good! It really cleans up the top of the notebook and stays out of the way when you don't need it, and is easy to use when updating the metadata.

garden_ai/notebook_metadata.py
```diff
@@ -221,6 +177,126 @@ def save_requirements_data(
...
+            disabled=False,
+        )
+    elif nb_meta.notebook_requirements.file_format == "conda":
+        req_widget = widgets.Textarea(
+            value=yaml.safe_dump(nb_meta.notebook_requirements.contents),
+            placeholder="Requirements",
+            layout=widgets.Layout(width="100%", height="300px"),
+            continuous_update=False,
+            disabled=False,
+        )
```
hholb, line 248: Nothing wrong with this, but a question for @OwenPriceSkelly: are we going to stop supporting conda? If so, we could probably leave this out.

garden_ai/notebook_templates/sklearn.ipynb
```diff
@@ -91,7 +59,7 @@
    },
    "outputs": [],
    "source": [
-    "my_hugging_face_repo = create_connector(\"https://hugginface.co/garden-ai/sklearn-iris\")"
+    "my_hugging_face_repo = create_connector(\"https://huggingface.co/garden-ai/sklearn-iris\")"
```
hholb, line 62: Good catch! Probably my fault - I have a knack for typos :laughing:

tests/test_metadata.py
hholb, line 1: Nice tests!

hholb (approved):
Everything looks good! I think this looks a lot better in the notebooks than the previous metadata cell, nice stuff!

garden_ai/notebook_metadata.py
```diff
@@ -221,6 +177,126 @@ def save_requirements_data(
...
+    )
+    metadata_widget = widgets.Accordion(
+        children=[doi_widget, base_image_widget, req_widget_tabs],
+        titles=("Global DOI", "Base Image", "Requirements"),
+    )
+
+    display(metadata_widget, output)
+
+
+def _pickle_metadata(nb_meta: NotebookMetadata):
```
OwenPriceSkelly, line 295: I really, really don't think we should be using pickle to persist anything we don't absolutely need to use pickle for. Because the data we're persisting here _can_ be JSON, it should be -- using pickle ties this metadata both to the python version used in the container and to the exact `NotebookMetadata` class definition we're using to read it, which is brittle. It would also be trivial to hand-build a valid .json for this but very hard to build a valid .pkl if any part of this breaks down.

OwenPriceSkelly (changes requested):
This is a very fun PR -- there's a lot that I _really_ like, a couple of questions, and just the one major gripe. 

side note: ended up with a lot of feedback here but I don't think the changes should take you very long. If it's looking like more than ~1 day to tie this up let me know; I don't want to totally kill momentum for your MDF work. This is fortunately a nice and self-contained PR so it's not the end of the world to revisit it a little later if you'd prefer (just to give you options). 

## things I like 
  - wayyyy better UX than the raw JSON-esque comment string. a little polish really does go a long way  
  - besides the widget itself being good UX, I think the on save hook is a very nice way to accomplish the under-the-hood behavior we want in a way that is unlikely (least likely, anyway) to be surprising to users. good stuff!  
## questions
  - why upgrade namedtuple -> dataclasses vs namedtuple -> pydantic models? your `as_dict` helpers especially remind me of pydantic `model_dump`s 
  - how do widget requirements and a `requirements.txt` specified in e.g a `garden-ai notebook start` command interact with each other? if I add a new requirement to the metadata via the widget, does that requirement get installed in my notebook?
## gripe
  - pickle is a _very_ brittle way to persist this metadata. since everything we're dealing with for this metadata is easily JSON-able, let's stick with JSON (more reasoning below, also happy to sync up and share some .pkl horror stories if you're curious lol) 
## other misc thoughts 
- like Will mentioned earlier today, I think a little copy-editing to clarify what the global DOI does is a good idea. Same goes for the requirements behavior I'm curious about 
- like Hayden mentioned in his review, we don't have to worry about conda. if you don't want to just delete all the conda-related stuff you've got already that's fine, but lets at least hide the conda-related options from users so we don't mislead them into thinking conda is present in the container.

garden_ai/containers.py
```diff
@@ -461,6 +474,26 @@ def build_image_with_dependencies(
...
 
+        # Add custom_jupyter_config.py script to image
+        # This is used to to add a post_save_hook to jupyter to also save notebooks metadata
+        import garden_ai.scripts.custom_jupyter_config  # type: ignore
+
+        script_jupyter_config_contents = inspect.getsource(
+            garden_ai.scripts.custom_jupyter_config
+        )
+
+        # Line causes error when trying to import, append after import to avoid
```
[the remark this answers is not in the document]
MaxTuecke: Its some black magic, i think its referencing the jupyter FileContentsManager. Looking into it a bit more, I should be able to import the FileContentsManager set it using that instead, so I should be able to remove that.

garden_ai/notebook_metadata.py
```diff
@@ -1,129 +1,107 @@
...
-            notebook_requirements,
+            ntbk["metadata"]["garden_metadata"]["global_notebook_doi"],
+            ntbk["metadata"]["garden_metadata"]["notebook_image_name"],
+            ntbk["metadata"]["garden_metadata"]["notebook_image_uri"],
+            RequirementsData(
+                **ntbk["metadata"]["garden_metadata"]["notebook_requirements"]
+            ),
         )
-    else:
+    except Exception:
```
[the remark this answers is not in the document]
MaxTuecke: Will change that to pydantic.ValidationError since I'm also changing NotebookMetadata and RequirementsData from dataclass to pydantic.BaseModel

garden_ai/containers.py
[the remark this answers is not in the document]
MaxTuecke: The workaround for not using `c` ended up being a complete pain and required some extra libraries. Im just going to leave as is and link to the docs.

MaxTuecke:
As we talked about in person, I added in a 'Install new requirements' button that pops up when any changes are made to the requirements w/ the widget. It will install new reqs, update the users local reqs file (if one was provided) and then restart the jupyter kernel.

As Owen / Hayden suggested I also:
- Changed `NotebookMetadata` and `RequirementsData` to pydantic BaseModel
- Changed the metadata saving from pickle to json
- Removed the conda functionality for the widget since we are planning on removing that functionality anyways.

OwenPriceSkelly (approved):
LGTM!! way to go the extra mile on this one, I'm a huge fan of where it's ended up. 

Also very clear that you spent a lot of time test-driving this, which I appreciate -- genuinely delighted to find that I couldn't break the widget by deleting the cell or moving it away from the top of the notebook. 

I've added suggestion comments with some small copy edits for typos/phrasing. The only real change I'm gonna ask before you merge is to not automatically restart the kernel when installing new requirements. More thoughts in comments, but having played around with it personally I think users might expect it to behave just like a `!pip install` in a regular cell, where the kernel restart is manual (if it's necessary). 

tl;dr this was a very nice execution on a great idea. we love to see it 💯

garden_ai/notebook_metadata.py
```diff
@@ -221,6 +195,254 @@ def save_requirements_data(
...
+            # remove update button from metadata_widget
+            new_children = list(metadata_widget.children)
+            new_children.remove(update_reqs_widget)
+            metadata_widget.children = new_children
+
+            # re-enable button
+            button.disabled = False
+
+            # restart jupyter kernel
+            IPython.Application.instance().kernel.do_shutdown(True)
```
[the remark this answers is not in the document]
MaxTuecke: Yeah, i agree it can defiantly look like a bug. Ill remove the kernel restart, add a popup msg for a successful install and put some help text there.

PR #495, opened 2024-07-03
local data toggle

### LINKED ISSUE #494: migration: toggle `local_data` module
Step 0 to replace our usage of local data with backend API calls is the ability to switch local_data off entirely when developing, without changing the behavior of the sdk for anyone else. 

In the presence of a `GARDEN_DISABLE_LOCAL_DATA` environment variable, the sdk should not be able to call anything defined in `local_data.py`.

OwenPriceSkelly:
closes #494 
## Overview

Very small PR to sanity check approach before actually replacing all of the local_data calls. This turns `import local_data` into an error in the presence of a special environment variable, and adds the basic CRUD methods to the backend client. 

## Discussion

There should be exactly zero behavioral difference for anyone without the `GARDEN_DISABLE_LOCAL_DATA` variable set except for one new (currently optional) `function_text` field on `RegisteredEntrypoint` for compatibility with the backend response schema. 

This is always the same as the function text otherwise stored in `.steps[0].function_text`, it's only necessary because the backend doesn't have steps.

## Testing
Without the variable I can `import garden_ai` just like before, but with the environment variable set it fails with flying colors. 

## Documentation

n/a

----
📚 Documentation preview 📚: https://garden-ai--495.org.readthedocs.build/en/495/

### DIFF
--- a/garden_ai/backend_client.py
+++ b/garden_ai/backend_client.py
@@ -6,7 +6,8 @@ import boto3
 import requests
 
 from garden_ai.constants import GardenConstants
-from garden_ai.gardens import PublishedGarden
+from garden_ai.gardens import Garden, PublishedGarden
+from garden_ai.entrypoints import RegisteredEntrypoint
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
LGTM!

Issue #496, opened 2024-07-03
migration: conditionally call local_data


OwenPriceSkelly:
Everywhere that calls a function defined in `local_data.py` should check the `local_data._IS_DISABLED` constant before it does. When that constant evaluates to true, the call site should use the appropriate `BackendClient` method instead.

PR #498, opened 2024-07-09
conditionally call local_data

OwenPriceSkelly:
completes #496 

## Overview

This PR uses the environment variable introduced in #495 to use the new backend as a drop-in replacement anywhere that currently uses `local_data`. 

## Discussion

This has some ugly hacks to keep everything consistent with current behavior, especially when it comes to the `Garden` vs `PublishedGarden` distinction. Some of those `Garden` helper methods are pretty crufty and I can't wait to get rid of them completely. 

We only even have the distinction there because of constraints due to local data in the first place -- one for normalized local_data entries, the other is for the denormalized search index entries. When the migration is through and the SDK no longer needs to worry about building the entries for the search index, we can simplify down to a single normalized `GardenMetadata` class. 

Until then, I suppose lazy-importing a client to talk to the backend is no more hacky than lazy-importing local_data to build the nested metadata for the search index. 

The only functionality we don't _quite_ have the backend ready to replace is the `list` subcommands, which just print/format all the gardens/entrypoints in local data. Once we have https://github.com/Garden-AI/garden-backend/issues/111 complete we can replace these with "list gardens that belong to me" functionality, but in the meantime those commands just aren't especially meaningful in a world without local_data, so now we just don't define them unless local_data is enabled. 

## Testing

Sanity checked all of the `garden-ai {garden, entrypoint, notebook}` subcommands exhaustively from the CLI with/without having the `GARDEN_DISABLE_LOCAL_DATA` variable set. If unset, everything works the same as before. If set, all the commands still work as expected but by calling the backend API instead of touching local_data. The only thing that still goes in local data is the user's email for the whoami command. 

## Documentation
no docs 

----
📚 Documentation preview 📚: https://garden-ai--498.org.readthedocs.build/en/498/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -503,12 +503,22 @@ class GardenClient:
 
             registered = RegisteredEntrypoint(**record)
             self._update_datacite(registered)
-            local_data.put_local_entrypoint(registered)
+            if local_data._IS_DISABLED:
+                self.backend_client.update_entrypoint(registered)
+            else:
+                local_data.put_local_entrypoint(registered)
 
             # fetch garden we're attaching this entrypoint to (if one was specified)
             garden_doi = metadata.get(f"{key}.garden_doi")
             if garden_doi:
-                garden = local_data.get_local_garden_by_doi(garden_doi)
+                if local_data._IS_DISABLED:
+                    published = self.backend_client.get_garden(garden_doi)
+                    garden = Garden(
+                        **published.model_dump(), _entrypoints=published.entrypoints
+                    )
+                else:
+                    garden = local_data.get_local_garden_by_doi(garden_doi)
+
                 if garden is None:
                     msg = (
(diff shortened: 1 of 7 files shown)


garden_ai/client.py
```diff
@@ -322,82 +321,6 @@ def _update_datacite(
         self.backend_client.update_doi_on_datacite(payload)
         logger.info("Update request succeeded")
 
-    def add_paper(self, title: str, doi: str, **kwargs) -> None:
```
OwenPriceSkelly, line 325: turned out we never actually used these two methods, so I just took them out

WillEngler: 🫡  Farewell

WillEngler (approved):
I think you shrewdly found the right "seam" here and did this in the least hacky way you could. Honestly this feels cleaner than I would have expected. Good stuff!

PR #503, opened 2024-07-19
Search index migration

### LINKED ISSUE #502: migration: flip the switch
Bookkeeping: this is for writing actual script itself and flipping the `_IS_DISABLED` flags for the next sdk release.

OwenPriceSkelly:
closes #502 
## Overview
At last! This is the script which re-populates the search index via the new backend API. It also hardcodes the `_IS_DISABLED` constants to `True` for the next release. 

## Testing

UPDATE: flipping the switch broke a lot of tests. Also, a significant portion of our test code just revolves around mocking local_data, our best-tested module by far. I spent some time fiddling with the broken tests, but the core problem of the `Garden` class lazy-importing local data (and/or a GardenClient while we're still in migration limbo) was making it far more effort than it's worth, especially when that's a hack we're about to remove entirely in #504. I also think that a "test suite overhaul" is frankly overdue for the sdk and I think it makes more sense to scope that out into its own story (perhaps for a testing-inclined intern?) than it does to block the migration until the tests have been reworked. But I'm open to pushback on this!

Manually tested on dev against both the actual dev API and a local instance of the API which points at the dev search index. I tested both because it was easier to wipe a the local postgres instance completely, which more accurately reflects the (still empty) prod DB than the dev DB does. 

As of https://github.com/Garden-AI/garden-frontend/issues/152 fixing the display issues this caused on the staging frontend, the migration was successful on dev 🥳 

## Discussion
Steps remaining before declaring final victory on the migration: 
  - deploy pending changes to prod backend that enable the search index syncing behavior and deprecate the old routes 
  - deploy prod frontend with schema fixes from staging 
  - cut new release of garden-ai package with the _IS_DISABLED flags set 
  - run this script with `GARDEN_ENV=prod` 

## Documentation
No docs 

----
📚 Documentation preview 📚: https://garden-ai--503.org.readthedocs.build/en/503/


WillEngler (approved):
I think splitting out the test suite rework makes sense - so much of the internals are totally different now. Good job and good luck on the flip switch!

Issue #507, opened 2024-07-25
test suite revamp


OwenPriceSkelly:
Parent issue for @hholb's sidequest of revamping the test suite for the SDK. High level goals we settled on in our recent sync, subject to change: 

1. comprehensive unit test coverage of the CLI commands themselves. These should still mock API calls and anything expensive like commands that require building an image.
2. end-to-end integration test (existing issue: #461) that exercises the CLI against our dev backend. This should run regularly and should include executing remotely on the demo endpoint, notifying us if there's a failure 
3. if not already accomplished for free with (1), target of 80% unit test coverage for the rest of sdk

PR #508, opened 2024-07-26
Owen/cleanup refactor

### LINKED ISSUE #504: migration: cleanup / refactor core classes
wrapping the migration is a good opportunity to clean out some cruft in the sdk 

- the local_data module itself can be removed 
- the concept of steps
- distinct `Garden` / `PublishedGarden` and `EntrypointMetadata` / `RegisteredEntrypoint` classes
  - an object structure that I think makes more sense than what we have now is to have pydantic classes responsible for metadata schemas (`EntrypointMetadata`, `GardenMetadata`), and regular classes (`Garden`, `Entrypoint`) which have a `metadata` attribute. The metadata classes would be data-only and responsible for mirroring the backend API schemas, and the regular classes would handle programmatic/interactive behavior, e.g. the implementation of things like `__dir__` or `__call__`, or building the datacite-specific json from the metadata etc. 
  - the important thing here is that any `Garden` methods which currently need to lazy-import local data or a garden client should be removed 
- redundant client methods like `delete_garden_from_search_index` vs `delete_garden_locally` 
- etc

### LINKED ISSUE #506: allow publishing "orphan entrypoints" from CLI
it's currently not possible to publish entrypoints that don't belong to a garden due to search index limitations, but after we flip the switch on the migration we can relax the constraint. This can be part of the release that switches off local data / deprecates the search index routes

OwenPriceSkelly:
closes #504 and closes #506 

## Overview

This was a pretty big refactor and rethinks the design of our core `Garden` and `Entrypoint` classes. I lost my original writeup to a poorly-timed browser refresh, so I synced with @WillEngler earlier talked him through the major changes. 

To summarize for bookkeeping: 
  - "mere metadata" classes are in the `schemas` submodule, and are only responsible for validating data entering or exiting the system. These are  `EntrypointMetadata`, `RegisteredEntrypointMetadata`, and `GardenMetadata`. The latter two correspond closely to the respective [backend schemas](https://api-dev.thegardens.ai/docs)
  - the "interactive" classes `Garden` and `Entrypoint` are responsible for things like `__call__`, `__dir__`, `_repr_html_` etc that allow users to invoke entrypoints from a garden like methods. these have a `metadata` attribute containing exactly one respective `*Metadata` object, which is required to instantiate. 
  

This PR also declares unit test bankruptcy, but we have test czar @hholb to bail us out in #507 and #511.

----
📚 Documentation preview 📚: https://garden-ai--508.org.readthedocs.build/en/508/


hholb:
:red_square: :red_square: :red_square: THAT DIFF :red_square: :red_square: :red_square:

garden_ai/local_data.py
```diff
@@ -1,235 +0,0 @@
-_IS_DISABLED = True
-
-
```
WillEngler, line 3: Farewell local_data 🫡

WillEngler (approved):
I really like the new class structure. This is fantastic, let's get it in!

garden_ai/local_data.py
OwenPriceSkelly: 🪦 RIP

Issue #515, opened 2024-08-07
Tests covering user-facing SDK modules


hholb:
Part of #507 

PR #514 added tests verifying the behavior of the CLI commands from a user's perspective.

We need similar tests covering the other user-facing areas of the SDK, namely:
- Model Connectors
- Garden Client
- Gardens
- Entrypoints

PR #517, opened 2024-08-12
Add small-scale tests of user-facing APIs

hholb:
Resolves: #515 
Related: #507 

## Overview

This PR add small-scale test coverage of the user-facing APIs in the SDK.

## Discussion

I looked through the tutorial and example notebooks to identify the primary functions and classes users will be interacting with. This PR adds test coverage of their expected behaviors from the user's perspective. All of these tests are small-scale meaning all network responses are mocked, and tests do not use docker.

The primary areas we are still lacking coverage are in the container and notebook modules. Test coverage for these modules will come in another PR.

In anticipation of adding some integration tests of the container and notebook modules, this PR adds a pytest marker: `@pytest.mark.integration` that should be added to all tests that use resources outside of the main garden-ai process, e.g. network services and docker. This PR also configures pytest to skip integration tests by default. Integration tests will run if you pass a `--integration` flag or `-m "integration"` to the pytest command.

## Testing

It's all tests!

Run the small-scale tests:
```
pytest
```

Run integration tests:
```
pytest --integration
# or
pytest -m "integration"
```

----
📚 Documentation preview 📚: https://garden-ai--517.org.readthedocs.build/en/517/

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,34 @@ from garden_ai.schemas.entrypoint import RegisteredEntrypointMetadata
 from garden_ai.schemas.garden import GardenMetadata
 
 
+def pytest_addoption(parser, pluginmanager):
+    parser.addoption(
+        "--integration",
+        action="store_true",
+        default=False,
+        help="run integration tests",
+    )
+
+
+def pytest_collection_modifyitems(session, config, items):
+    if config.getoption("--integration"):
+        # if --integration is set, don't skip integration tests
+        return
+
+    if marker := config.getoption("-m"):
+        if "integration" in marker and "not integration" not in marker:
+            # if -m "integration" is given on the cli, don't skip integration tests
+            return
+
+    # Otherwise, skip integration tests
+    skip_integration = pytest.mark.skip(
+        "need -m 'integration' or --integration option to run"
+    )
+    for item in items:
+        if "integration" in item.keywords:
+            item.add_marker(skip_integration)
+
+
 @pytest.fixture
 def cli_runner() -> CliRunner:
     """Return a typer.testing.CliRunner for use in tests."""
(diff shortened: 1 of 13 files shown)


MaxTuecke (approved):
LGTM! All tests passing on my machine. TY for picking up the notebook metadata stuff

tests/conftest.py
```diff
@@ -188,3 +220,71 @@ def mock_RegisteredEntrypointMetadata(
 ) -> RegisteredEntrypointMetadata:
     """Return a RegisteredEntrypointMetadata object populated with test data."""
     return RegisteredEntrypointMetadata(**entrypoint_metadata_json)
+
+
+@pytest.fixture
```
OwenPriceSkelly, line 225: I haven't seen the faker library before but now I'm intrigued, this looks very handy

OwenPriceSkelly: reminds me of another testing library I've never used but seems like a fun rabbit hole: https://hypothesis.readthedocs.io/en/latest/manifesto.html

tests/model_connectors/test_github_conn.py
```diff
@@ -0,0 +1,143 @@
...
+    ), "Pull should be called on subsequent calls"
+
+
+def test_GitHubConnector_pins_commit_hash_correctly(
+    mocker,
+    patch_has_lfs,
+    patch_fetch_readme,
+):
+    # Mock the response from the GitHub API
+    fake_commit_hash = "a" * 40
```
OwenPriceSkelly, line 81: every time I see this I read it in my head like the first bit of immigrant song lol

tests/test_entrypoints.py
```diff
@@ -0,0 +1,265 @@
...
+    result = normal_entrypoint("test arg", endpoint="specific-endpoint")
+
+    assert (
+        result == "mocked result"
+    ), "Should return the direct result for an entrypoint with DOI not in list"
+    mock_executor_instance.submit_to_registered_function.assert_called_once()
+
+
+def test_entrypoint_test_raises_on_non_idempotent_entrypoint(mocker):
+    class NonIdempotentCounter:
```
OwenPriceSkelly, line 100: just an understanding check bc I noticed this one is a bit longer than the others -- was there something especially tricky to mock/test here that we may want to revisit design-wise? or is this one just long because it's thorough, and wouldn't have benefit from splitting into smaller tests

OwenPriceSkelly (approved):
LGTM! I salute the test czar 🫡

PR #524, opened 2024-08-20
add get_email call to __init__

### LINKED ISSUE #523: bug: it's possible to skip joining the "Garden Users" group
During the workshop KJ led last week, a few people ran into auth errors trying to run on the demo endpoint. This is because the sdk was submitting a job directly to the endpoint before they had authed with our backend, so they weren't added to the group automatically.

OwenPriceSkelly:
fixes #523 
## Overview

This just adds a single get_email call to the `GardenClient.__init__` method, which ensures that the client has authed with our backend at least once before attempting to pull a garden or entrypoint to run on the demo endpoint. 

This makes it so you can't accidentally try running an entrypoint before the backend has had the chance to add you to the Garden Users globus group. 

## Discussion
This is definitely a bit of a hack, but so is the backend's automatically adding users to the group. If there's an obvious cleaner way to do this I'm open to it, this was just the first easy fix to come to mind.

## Testing
I had to tweak the garden_client fixture so the call to `BackendClient.get_user_info` would be mocked with test user data. This fixed almost all of the tests except for one backend client test, since the `backend_client` fixture depends on the `garden_client` fixture (which now mocks the backend client method being tested). 

I'll defer to the judgement of the test czar @hholb, but I decided it was ok to delete the single remaining affected test. It seemed like the logic it was testing of "an exception is raised by the backend" is something that could be (and already is, afaik) covered by the backend's unit tests, but I'm open to pushback on this!

Another more robust / time consuming option would be to make the `backend_client` fixture independent of the `garden_client` fixture, I just didn't think it was worth the effort for this PR

----
📚 Documentation preview 📚: https://garden-ai--524.org.readthedocs.build/en/524/

### DIFF
--- a/garden_ai/client.py
+++ b/garden_ai/client.py
@@ -30,6 +30,7 @@ from globus_sdk.scopes import ScopeBuilder
 from globus_sdk.tokenstorage import SimpleJSONFileAdapter
 from rich import print
 from rich.prompt import Prompt
+
 from garden_ai.backend_client import BackendClient
 from garden_ai.constants import GardenConstants
 from garden_ai.entrypoints import Entrypoint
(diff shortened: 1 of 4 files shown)


WillEngler (approved):
Cool, let's do it!

PR #528, opened 2024-08-21
docs revamp

### LINKED ISSUE #512: give the docs some ✨ polish ✨
Some gripes about the docs: 

- Clean up home page ("Docs are a WIP" get out of here)
  - 1-sentence elevator pitch and TOC is good
  - what's the motivation to use this/read the rest of the docs?
- Explain Entrypoint concept before Garden concept 
- Should link to the website
  - should link to real examples on the website 
- Fix API autodocs for metadata classes
  - Include usage example for linking Dataset/Paper/Models 
- see if we can drop the `garden-ai` prefix on the CLI autodocs TOC
- the "User guide" > "Introduction" page is almost identical to the home page 
- Flesh out FAQs page

OwenPriceSkelly:
closes #512 

## Overview

This PR cleans up our docs in a few places, with the major new addition being a (hopefully) helpful API reference for the main user-facing classes in the SDK (Client, Gardens, Entrypoints, *Metadata objects, plus decorators and helpers). 

The "automatic" documentation also required some painstaking updates to the docstrings for the classes I wanted to document, so the diff isn't entirely limited to the `docs/` directory (no functional changes though). This also involved prefixing some methods with underscores so they didn't get pulled into the generated docs. 

I also changed the decorator name from `garden_entrypoint` to `entrypoint`, but I kept `garden_entrypoint` around as an alias for backwards-compatibility 

## Discussion

I still have some lingering gripes with the docs, but I don't want to spend more much more time on the docs right now until we have the users to read them. But I think this reaches the short term goal of "a hypothetical user could work out how to link their paper or dataset metadata to their entrypoints entirely from the docs" which I'm happy with. 

If we don't think that goal is met yet I'm happy to revisit, but I think other docs gripes can wait / fall under their own PR 

## Testing

with `mkdocs serve` and my own two eyes 👀 

## Documentation

All docs!

----
📚 Documentation preview 📚: https://garden-ai--528.org.readthedocs.build/en/528/


WillEngler (approved):
Woo! We love docs

Issue #521, opened 2024-08-19
Can't use model connector with GitHub


WillEngler:
To repro ...

```
from garden_ai.model_connectors import create_connector
con = create_connector('https://github.com/WillEngler/model-perovskite-ASR')
con.stage()
```

I see ...

```
Traceback (most recent call last) ────────────────────────────────╮
│ /usr/local/lib/python3.11/site-packages/garden_ai/model_connectors/model_connector.py:156 in     │
│ stage                                                                                            │
│                                                                                                  │
│   153 │   │   │   ConnectorStagingError: when something goes wrong during staging.               │
│   154 │   │   """                                                                                │
│   155 │   │   try:                                                                               │
│ ❱ 156 │   │   │   self._checkout_revision()                                                      │
│   157 │   │   │                                                                                  │
│   158 │   │   │   # pull from the repo if we have already downloaded it.                         │
│   159 │   │   │   if self._pull_if_downloaded():                                                 │
│                                                                                                  │
│ /usr/local/lib/python3.11/site-packages/garden_ai/model_connectors/github_conn.py:104 in         │
│ _checkout_revision                                                                               │
│                                                                                                  │
│   101 │   │   │   ValueError: when there is an error checking out the revision.                  │
│   102 │   │   """                                                                                │
│   103 │   │   try:                                                                               │
│ ❱ 104 │   │   │   repo = Repo(self.local_dir)                                                    │
│   105 │   │   │   repo.git.checkout(self.revision)                                               │
│   106 │   │   except GitCommandError as e:                                                       │
│   107 │   │   │   raise ConnectorRevisionError(                                                  │
│                                                                                                  │
│ /usr/local/lib/python3.11/site-packages/git/repo/base.py:236 in __init__                         │
│                                                                                                  │
│    233 │   │   epath = expand_path(epath, expand_vars)                                           │
│    234 │   │   if epath is not None:                                                             │
│    235 │   │   │   if not os.path.exists(epath):                                                 │
│ ❱  236 │   │   │   │   raise NoSuchPathError(epath)                                              │
│    237 │   │                                                                                     │
│    238 │   │   # Walk up the path to find the `.git` dir.                                        │
│    239 │   │   curpath = epath                                                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
NoSuchPathError: /garden/models/model-perovskite-ASR

The above exception was the direct cause of the following exception:

[...]

PR #529, opened 2024-08-22
fix: staging error in GitHubConnector

hholb:
Resolves: #521

# Overview
This PR fixes a bug @WillEngler found when calling `stage` on a `GitHubConnector`.

# Discussion
The error was due to calling `_checkout_revision` before cloning/downloading the repo. I moved the call to `_checkout_revision` to after the download step in `stage` to ensure the repo is present before attempting to checkout the tagged revision.

# Tests
Added a integration test covering the logic Will reported in #521
Run with:
`pytest -m "integration"`

----
📚 Documentation preview 📚: https://garden-ai--529.org.readthedocs.build/en/529/


OwenPriceSkelly (approved):
LGTM!

PR #530, opened 2024-08-22
Add small-scale tests of core modules

hholb:
Related: #507 

## Overview

This PR add small-scale test coverage of the the core modules that make garden work.

## Discussion

This PR includes a sizeable refactor of the `notebook_metadata.py` module to make it easier to get better test coverage. I didn't change any of the underlying logic, mostly moved stuff around into helper functions.

The tests of the containers module are mostly from the old test suite. The containers module and notebook CLI commands are where the biggest gaps in coverage are after this PR. Integration tests that cover these modules will come in another PR.

## Testing

Lots of new tests, plus some old ones!

----
📚 Documentation preview 📚: https://garden-ai--530.org.readthedocs.build/en/530/

### DIFF
--- a/garden_ai/notebook_metadata.py
+++ b/garden_ai/notebook_metadata.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional, Union, Callable
 from pydantic import BaseModel, ValidationError
 import typer
 import json
@@ -67,10 +67,10 @@ def add_notebook_metadata(
     # If metadata widget cell does not exist, add to top of notebook
     if not _has_metadata_cell_tag(ntbk):
         new_cell = nbformat.v4.new_code_cell(NOTEBOOK_DISPLAY_METADATA_CELL)
+        del new_cell["id"]
         new_cell["metadata"] = {
             "tags": [METADATA_CELL_TAG],
         }
-        del new_cell["id"]
         ntbk.cells.insert(0, new_cell)
 
     # Add empty garden_metadata dict to notebooks metadata if missing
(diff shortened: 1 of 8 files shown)


garden_ai/notebook_metadata.py
OwenPriceSkelly, line 1: I'll defer to @MaxTuecke's LGTM on the refactor but at a glance this looks sane to me!

tests/conftest.py
```diff
@@ -288,3 +305,129 @@ def patch_infer_revision(mocker):
...
+
+@pytest.fixture
+def pip_requirements_raw() -> str:
+    """Return a string of pip requirements a la requirements.txt"""
+    return "scikit-learn==1.2.2\npandas\n"
+
+
+@pytest.fixture
+def pip_requirements() -> dict:
+    """Reeturn a dict with structure of a RequirementsData object"""
```
OwenPriceSkelly, line 338: 👀 🔍 
```suggestion
    """Return a dict with structure of a RequirementsData object"""
```

tests/conftest.py
```diff
@@ -288,3 +305,129 @@ def patch_infer_revision(mocker):
     )
 
     yield
+
+
+@pytest.fixture
+def backend_client(garden_client):
```
OwenPriceSkelly, line 311: understanding check - what's the difference between the BackendClient returned by this fixture and the one at `garden_client.backend_client`?

OwenPriceSkelly (approved):
LGTM! 🙌 

Also, FWIW I think the notebook_metadata refactor makes sense in the context of this PR but would also have been totally fair to split out into its own issue if it felt out of scope

tests/conftest.py
hholb: There isn't a difference functionally. I think it makes sense in `test_backend_client.py` to have a BackendClient to test directly rather than having to use `garden_client.backend_client....` in each test (even if that is essentially what happens in the fixture).

PR #533, opened 2024-09-12
Create an end to end test of the user tutorial

hholb:
#507 

## Overview

This PR adds an end-to-end test of the garden user tutorial.

Pytest is configured to skip ete tests by default. Ete tests should be marked with `@pytest.mark.ete`.
Ete test can be run with:
```bash
pytest -m "ete"
```
Individual ete tests can be run filtering by name. For example to run the tutorial ete test:
```bash
pytest -m  "ete" -k "tutorial"
```

## Discussion

I created an `ete` module in the `tests/` directory that has its own `conftest.py` setting up fixtures specific for end-to-end tests. The idea is to implement ete tests as a chain of fixtures that encapsulate discrete steps in various flows through garden. This allows an each step to do some work, yield the result to the next step, and then clean up after the test is finished. The final test case is a regular pytest test function that depends on one or more fixtures and makes some assertions about the state of world. Once the test is finished the fixtures can 'unwind' and clean up after themselves.

The `create_garden` fixture in `tests/ete/conftest.py` is a good place to start looking.
See `tests/ete/test_tutorial.py` to get an idea of how the whole system works.

In theory, this should allow us to write new ete tests easily by reusing fixtures that encapsulate one part of a users path through different garden functionality.

## Testing

It is a test!

Lots of manual testing.

----
📚 Documentation preview 📚: https://garden-ai--533.org.readthedocs.build/en/533/


OwenPriceSkelly (approved):
LGTM! I salute the test czar 🫡

PR #536, opened 2024-09-26
invoke modal function from sdk

### LINKED ISSUE #534: Update SDK to incorporate Modal Functions
Gardens will soon contain Modal Functions in addition to Entrypoints. The SDK will need to be able to parse these new entities and allow them to be called as functions.

Actually invoking the Modal Functions will depend on a POST /modal-function-invocations backend endpoint. So it might make sense for this PR to just add the Pydantic schema and the function invocation can be a no-op.

# Assumptions:
1.
2.

# Acceptance Criteria
~Given I am logged in with the SDK, when I fetch a garden that has Modal Functions, then the garden SDK object will have all the modal functions available and callable.~ (now #537) 

Given a `ModalFunctionMetadata` with the real name of a modal app / function + a `ModalFunction` object with that metadata, the `ModalFunction` object is callable and hits the backend `POST /modal-function-invocations` route.

OwenPriceSkelly:
closes #534
## Overview

This PR has a first pass of a callable `ModalFunction` object that can call the new POST /modal-invocations route on the backend (🥳)

## Discussion
This introduces a handful of new classes for requesting modal function invocation on the backend. There's also a new `BackendClient.invoke_modal_function` method, which does what it says on the tin. 

### New Schemas

I copied the `ModalInvocationRequest`/`Response` schemas from the corresponding backend PR and stubbed out a `ModalFunctionMetadata` schema, which currently only has the two fields necessary to invoke the function but will eventually include the other metadata fields we persist in the DB (so will look more like the `EntrypointMetadata` schema) 

The one ~fancy thing to highlight about the schemas (also borrowed from the backend PR) is the custom pydantic type `B64Bytes` for the serialized argument/response payloads. As far as type checking / pydantic / your IDE is concerned this is the same as regular `bytes`, but when translating to/from JSON it will base64 encode the bytes so that it's a plain string. 

### New Object 

The `ModalFunction` object has the same relationship to `ModalFunctionMetadata` as `Entrypoint` does to `RegisteredEntrypointMetadata`. You need metadata to init the object, then calling the object will handle making a request per that metadata. 

## Testing

Manually against a local backend instance, like in the demo: 
```python
### modal-demo.py

def main():
    from garden_ai.schemas.modal import ModalFunctionMetadata
    from garden_ai.modal.functions import ModalFunction

    metadata = ModalFunctionMetadata(app_name="hello-modal", function_name="hello_garden")
    modal_function = ModalFunction(metadata=metadata)

    x = "I'm an input argument!"
    return modal_function(x)


if __name__ == "__main__":
    print(main())
```
pointed at a local instance of the backend:
`GARDEN_ENV=local python -m modal-demo`

For unit tests I added a happy test and a ~scary test (it's october) to make sure that the 2MiB argument size limit raises the right exception. 

## Documentation

no docs!

----
📚 Documentation preview 📚: https://garden-ai--536.org.readthedocs.build/en/536/


pyproject.toml
```diff
@@ -43,7 +43,7 @@ jinja2 = "^3.1.2"
 huggingface-hub = "0.18.0"
 pyyaml = "^6.0"
 packaging = "^23.0"
-globus-compute-sdk = "^2.2.0"
```
OwenPriceSkelly, line 46: This was because a change to the globus compute client attribute names was breaking our tests, but outside of redoing how we mock the client I don't think there'd be any issues with 2.28.0

tests/modal/test_functions.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Check that _modal_process_result_sync was called
+    mock_modal_process_result_helper.assert_called_once()
+
+    # Check that the result was processed correctly
+    assert result == "what results! and *so* nicely processed"
+
+
+def test_modal_function_size_limit(modal_function):
+    scary_input = b"oo" * (1024 * 1024 + 1)  # just over 2MiB
```
WillEngler, line 85: good October variables

WillEngler (approved):
LGTM! Let's keep this train rolling 🚂

PR #539, opened 2024-10-03
gardens have ModalFunction attributes 

### LINKED ISSUE #537: feat: `Garden` objects have `ModalFunction` attributes (like `Entrypoints` currently)
Given I am logged in with the SDK, when I fetch a garden that has Modal Functions, then the garden SDK object will have all the modal functions available and callable.

This should complete the loop following #536, which introduces the callable `ModalFunction` objects but does not relate them to gardens

OwenPriceSkelly:
closes #537 

## Overview

Follow up to #536, this gives `Garden` objects a `modal_functions` attribute as well as the ability to call those functions by name, so the interface is the same as calling normal entrypoints. 

## Discussion

The implementation is almost identical to the way that Garden objects look up and call their entrypoints, but note that there are a few TODOs sprinkled here and there for future PRs. 

Specific point of divergence: gardens refer to entrypoints by DOI, but we don't have DOIs for modal functions (and probably won't for a while). I decided to go with the `function_name` in lieu of a doi for now, but depending on when we get the backend side of this feature (https://github.com/Garden-AI/garden-backend/pull/165) merged it's probably easier to just use the id from the database or something. 

Also, bad news: there is a circular import that I'm not sure we can easily fix without compromising the callable interface (`client` -> `backend_client` -> `gardens` -> `modal.functions` -> `client`). 

Good news is that ol' reliable (lazy importing `GardenClient` in `ModalFunction.__init__`) works fine as a short term fix, but we should keep in mind that every time we use this technique to dodge a circular dependency, we're actually dealing 2d6 psychic damage to our future selves. Life is all about tradeoffs. 

## Testing

I have an [updated version of last week's demo](https://gist.github.com/OwenPriceSkelly/b42dcfc308e87fc30c16273e279b7457) working (🥳) 

Also added a couple new tests / some fixture data, but mostly just modified existing tests to cover the new attributes `Garden`s have now. 

## Documentation

No docs! 

----
📚 Documentation preview 📚: https://garden-ai--539.org.readthedocs.build/en/539/

### DIFF
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -83,7 +83,7 @@ class Garden:
 
         for modal_function in self.modal_functions:
             if name == modal_function.metadata.function_name:
-                return name
+                return modal_function
 
         raise AttributeError(
             f"'{self.__class__.__name__}' object has no attribute '{name}'."
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -160,10 +160,13 @@ class Garden:
             entrypoints += [Entrypoint(RegisteredEntrypointMetadata(**entrypoint_data))]
             metadata.entrypoint_ids += [entrypoint_data["doi"]]
 
-        modal_functions = []
-        for modal_fn_data in data["modal_functions"]:
-            modal_functions += [ModalFunction(ModalFunctionMetadata(**modal_fn_data))]
-            # TODO: use DOIs once modal functions have them
-            metadata.modal_function_ids += [modal_fn_data["function_name"]]
+        if "modal_functions" in data:
+            modal_functions = []
+            for modal_fn_data in data["modal_functions"]:
+                modal_functions += [
+                    ModalFunction(ModalFunctionMetadata(**modal_fn_data))
+                ]
+                # TODO: use DOIs once modal functions have them
+                metadata.modal_function_ids += [modal_fn_data["function_name"]]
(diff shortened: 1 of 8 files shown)


garden_ai/gardens.py
```diff
@@ -62,6 +87,10 @@ def __getattr__(self, name):
             elif name == short_name:
                 message_extra = f" Did you mean {alias}?"
 
+        for modal_function in self.modal_functions:
+            if name == modal_function.metadata.function_name:
+                return modal_function
```
WillEngler, line 92: Something I'm wondering is how we'll handle methods on Modal classes. I think the backend should handle this by just storing the dotted function name (eg, `my_class.my_method`) but that will probably be weird for the SDK.

WillEngler (commented):
Looking good, just dashing out a note

WillEngler (approved):
LGTM!

garden_ai/gardens.py
OwenPriceSkelly: talked about this on a call but for posterity: I think that we'll have a `ModalClass` analogue to `ModalFunction` (so Gardens will have entrypoints, modal functions, and modal classes) and a `ModalClass.__getattr__` implementation *should* get us the intuitive syntax of `my_garden.my_modal_class.my_modal_method` without much extra legwork

PR #541, opened 2024-10-11
Remove search command and references to the search index

### LINKED ISSUE #538: Remove uses of the search index from the SDK
To finish weaning ourselves off the search index, we need to make sure the SDK/CLI isn't querying it. We think this is just happening in the `garden-ai` search command.

# Assumptions:
1. Whoever picks up this PR should also do a more thorough sweep of the SDK to ensure that there really is only one place it's calling search.

# Acceptance Criteria
Given I am a Garden SDK/CLI user, when the Garden search index is turned off, then I don't notice a thing - everything functions as it used to.

WillEngler:
Closes #538

## Overview

This PR removes the `garden-ai garden search` command and all references to the search index in the CLI. (With the exception of the search index migration script)

## Discussion

At first I was going to convert to the new search endpoint but ... I don't think we really need this feature in the CLI. It was one of the very first things we added to the CLI and it was largely showing that we had it hooked up to the search index. It just prints JSON. I think users can and should just use the website.

## Testing

I ran `garden-ai garden` and confirmed that the `search` subcommand was gone.

## Documentation

I removed the command from the docs

----
📚 Documentation preview 📚: https://garden-ai--541.org.readthedocs.build/en/541/

### DIFF
--- a/docs/garden-ai.md
+++ b/docs/garden-ai.md
@@ -229,7 +229,6 @@ $ garden-ai garden [OPTIONS] COMMAND [ARGS]...
 * `list`: Lists all local Gardens.
 * `publish`: Push data about a Garden stored to Globus...
 * `register-doi`: Moves a Garden's DOI out of draft state.
-* `search`: Queries the Garden search index and prints...
 * `show`: Shows all info for some Gardens
 
 ### `garden-ai garden add-entrypoint`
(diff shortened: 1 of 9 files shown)


OwenPriceSkelly (approved):
LGTM! 

so long, search index 🫡 🫗

PR #542, opened 2024-10-11
update schemas to use function IDs

### LINKED ISSUE #540: use IDs instead of function name for modal functions
small follow up ticket to #539 to update schemas etc and test once the respective [backend](https://github.com/Garden-AI/garden-backend/issues/166) [PRs](https://github.com/Garden-AI/garden-backend/pull/165) are in

OwenPriceSkelly:
## Overview
This is a small second pass PR which closes #540 and is basically just limited to schema tweaks on the sdk side. Nothing is functionally new vs #539. 

see also: the [corresponding backend PR](https://github.com/Garden-AI/garden-backend/pull/177) which expects the client to refer to functions by ID instead of function and app name.

~If you have this branch checked out for the sdk and `owen/post-modal-fns-uses-ids` checked out for a local instance of the backend, [this](https://gist.github.com/OwenPriceSkelly/57fa0abbddb6f63aaf4a42da205cd06a) script should reproduce the bug causing modal's containers to crash~

## Discussion

I decided I didn't like our custom UniqueList type throwing errors when it could just ignore repeat values and ensure the list was unique itself. This is more in line with pydantic's general philosophy of "parse don't validate even though we call it validate". 

This change wasn't prompted by anything in particular; just felt like being a good boyscout 

## Testing

I had to tweak a couple of unit tests but mostly just updating the fixture data was enough for everything to pass again. Also tested pretty extensively with my end-to-end script I originally linked. 
## Documentation

No docs 

----
📚 Documentation preview 📚: https://garden-ai--542.org.readthedocs.build/en/542/

### DIFF
--- a/garden_ai/gardens.py
+++ b/garden_ai/gardens.py
@@ -58,9 +58,8 @@ class Garden:
                 "Expected `entrypoints` DOIs to match `metadata.entrypoint_ids`. "
                 f"Got: {[ep.metadata.doi for ep in entrypoints]} != {metadata.entrypoint_ids}"
             )
-        # TODO: update to use DOIs when modal functions have them
         if set(metadata.modal_function_ids) ^ set(
-            [mf.metadata.function_name for mf in modal_functions]
+            [mf.metadata.id for mf in modal_functions]
         ):
             raise ValueError(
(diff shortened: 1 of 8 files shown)


WillEngler (approved):
LGTM!

Issue #552, opened 2024-11-12
Integrate async Modal invocation with SDK


WillEngler:
A number of tickets will make it possible to execute Modal functions asynchronously via the API. We need to take advantage of this from the SDK.

# Assumptions:
1. This will work similarly to how we currently run Globus Compute functions - from the SDK caller's perspective this will be a blocking call. Behind the scenes the SDK will grab a job id and poll for it to be done.
2.

# Acceptance Criteria
Given I am trying to execute a Modal function that takes more than 30 seconds to execute, when I execute it, then it runs and gives me a result instead of timing out.

PR #553, opened 2024-11-18
Handle modal invocations asynchronously

hholb:
Resolves: #552 

## Overview

This PR updates the backend client to hit the new async modal invocation routes. The invocations are still synchronous from the user's perspective, but this allows for long-running invocations without hitting the gateway timeout for a single request to the backend.

 There are no major changes in logic other than a polling loop to check the status of the invocation. There is a new environment variable "BACKEND_POLL_INTERVAL_SECONDS" to configure how often we poll for the invocation status. The default is set for 0.1 seconds, or 100ms.

## Discussion

We may want some sort of back-off logic in the poll interval so really long running jobs don't ping the server so much, but this probably won't be an issue until we have hundreds of invocations in-progress simultaneously. I'll leave that for an issue in the future.

## Testing
Updated existing tests.
Manual testing.

## Documentation
Nothing changes from the users perspective.

----
📚 Documentation preview 📚: https://garden-ai--553.org.readthedocs.build/en/553/

### DIFF
--- a/.safety-policy.yml
+++ b/.safety-policy.yml
@@ -4,89 +4,91 @@
 # Otherwise, you can use the flag `safety check --policy-file <path-to-this-file>` to specify a custom location and name for the file.
 # To validate and review your policy file, run the validate command: `safety validate policy_file --path <path-to-this-file>`
 security: # configuration for the `safety check` command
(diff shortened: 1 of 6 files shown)


WillEngler (approved):
LGTM!

PR #554, opened 2024-11-19
Modal class support

WillEngler:
Second part of #198. Closes #198

## Overview

This PR accommodates Modal class methods. They come over as metadata just as normal functions do - they just have dots in their names, like "MyClass.my_method". The SDK will construct wrappers on the fly so that users can invoke class methods like my_garden.MyClass.my_method

## Testing

Added a new unit test to exercise the chain of `__getattr__`s. Also did manual testing to check tab completion and HTML rendering.

Manual testing was limited as I mostly used fixture data.

## Documentation

Not yet, but this will definitely merit some mention in the docs

----
📚 Documentation preview 📚: https://garden-ai--554.org.readthedocs.build/en/554/

### DIFF
--- a/garden_ai/modal/classes.py
+++ b/garden_ai/modal/classes.py
@@ -34,6 +34,12 @@ class ModalClassWrapper:
             return self._methods[method_name]
         raise AttributeError(f"'{self.class_name}' has no method '{method_name}'")
 
+    def __dir__(self) -> list[str]:
+        """Return list of valid attributes for tab completion."""
+        return list(super().__dir__()) + list(  # Get standard attributes
+            self._methods.keys()
+        )  # Add all method names
+
     @classmethod
     def from_metadata(
         cls,
--- a/garden_ai/modal/classes.py
+++ b/garden_ai/modal/classes.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING, TypeVar, Any
 from garden_ai.modal.functions import ModalFunction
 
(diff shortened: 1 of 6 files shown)


garden_ai/modal/classes.py
```diff
@@ -0,0 +1,56 @@
...
+        class_name: str,
+        methods: list[ModalFunction],
+        client: GardenClient | None = None,
+    ):
+        self.class_name = class_name
+        # Create method lookup internally
+        self._methods = {
+            method.metadata.function_name.split(".")[-1]: method for method in methods
+        }
+        self.client = client
```
OwenPriceSkelly, line 29: would instances of this really need a client to `__init__`? since calling a modal class will resolve to a `ModalFunction` attribute (which already has a client) anyways, I think the only place we really need a client arg is the `from_metadata` method so we can inject one for the `ModalFunction`s

garden_ai/modal/classes.py
```diff
@@ -0,0 +1,56 @@
...
+            self._methods.keys()
+        )  # Add all method names
+
+    @classmethod
+    def from_metadata(
+        cls,
+        class_name: str,
+        methods_metadata: list[ModalFunctionMetadata],
+        client: GardenClient | None = None,
+    ) -> "ModalClassWrapper":
```
OwenPriceSkelly, line 49: style nit: if we `from __future__ import annotations` at the top this could just be the actual type instead of a string 
```suggestion
from __future__ import annotations
...
...
    ) -> ModalClassWrapper:
```

OwenPriceSkelly (approved):
LGTM! couple tiny nits but nothing to block over 🚢🫡

PR #561, opened 2025-01-15
hotfix: cap demo globus compute endpoint version 

OwenPriceSkelly:
PR just for bookkeeping, fix is live

## Overview

This fixes the mysterious "serialization" bug that broke the demo endpoint recently. Turned out not to have anything to do with serialization at all 🤷 

## Discussion 

It was a (kind of) version mismatch after all, just between the endpoint manager pod and the worker pods, not between the garden-side invocation and the worker pod. 

The manager pod always runs in the `funcx/kube-endpoint` image, while worker pods re-install globus-compute-endpoint on init. But the `funcx/kube-endpoint` image stopped receiving updates sometime last year (circa globus-compute-endpoint v2.16), which meant that the versions on either end of the "interchange" process started drifting, and eventually (around thanksgiving - v2.31) a change of signature in a helper function broke the camel's back. 

Because we're the only people still using a kube endpoint, I think we're the only people this could have happened to (everyone else could use the same install for both sides of the manager-worker interchange afaik) 

## Testing:

Screenshot 2025-01-14 at 7 13 41 PM

## Documentation
n/a

----
📚 Documentation preview 📚: https://garden-ai--561.org.readthedocs.build/en/561/


WillEngler (approved):
LGTM, thank you!!

Issue #566, opened 2025-01-21
bug: flaky entrypoint unit tests


OwenPriceSkelly:
A couple of unit tests (both from test_entrypoint) are flaky - if I am not connected to the internet, the `test_garden_entrypoint_decorator` test fails, and occasionally the `test_get_entrypoint_get_entrypoint_data_from_backend` fails in our CI build ([example](https://github.com/Garden-AI/garden/actions/runs/12892467554/job/35946735672#step:10:119)), but I can't reproduce this failure locally.

PR #569, opened 2025-01-28
Fix: flakey tests

hholb:
Resolves: #566 

## Overview

Fixes the flakey test issues. I didn't find the root cause of the flakey backend test, so I opted to remove it. The test was not very useful (my doing), and the behavoir is covered by other tests in the suite. Removal seems to be the best option.

The test that was failing when there was no internet access was due to a bug in the fixture mocking the calls to hugging face when creating a model connector. Easy fix!

### DIFF
--- a/garden_ai/entrypoints.py
+++ b/garden_ai/entrypoints.py
@@ -145,11 +145,6 @@ class Entrypoint:
         )
         return style + title + details + optional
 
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
Love the pragmatism here. LGTM!

PR #574, opened 2025-03-28
Add module level get_garden, get_functions funcs

hholb:
## Overview

This PR adds top-level `get_garden` and `get_function` functions to the SDK. This allows users to do this:

```python
import garden_ai

garden = garden_ai.get_garden("some/doi")
result = garden.run_some_function(inputs)
```

Instead of:
```python
import garden_ai
gc = GardenClient()
garden = gc.get_garden()
result = garden.run_some_function(inputs)
```

The functions are simple and just wrap the `gc = GardenClient()` up to remove that step. reducing the number of things our users need to type to invoke a Garden.

## Testing

Manually

## Documentation

If we like this, we should probably update the docs and examples to use these new functions.

### DIFF
--- a/garden_ai/__init__.py
+++ b/garden_ai/__init__.py
@@ -26,3 +26,13 @@ __all__ = [
     "entrypoint_test",
     "create_connector",
 ]
+
+
+def get_garden(doi: str) -> Garden:
+    gc = GardenClient()
+    return gc.get_garden(doi)
+
+
+def get_function(doi: str) -> Entrypoint:
+    gc = GardenClient()
+    return gc.get_entrypoint(doi)


OwenPriceSkelly (approved):
I like this a lot, big fan of cutting out the client initialization step. extremely cool that we have it down to a possible minimum of 3 lines of code to run remote inference (counting the import) 

  - I think we probably want to add `"get_garden"` to the `__all__` list
  - I think we should skip the `get_function` variant for now (or rename to `get_entrypoint`, but imo skipping is better), since that might be confusing
  - a small docstring example would be nice too, and could mention that it will prompt the user to login with globus if they need to. 
  
otherwise LGTM!

PR #577, opened 2025-04-16
Remove references to PUT /gardens

WillEngler:
This is a cleanup that is part of https://github.com/Garden-AI/garden-backend/issues/285. The SDK is the only place where PUT /gardens is called, and it is no longer in use.

----
📚 Documentation preview 📚: https://garden-ai--577.org.readthedocs.build/en/577/


scripts/migrate_search_index.py
OwenPriceSkelly, line 1: do we still need this script? i think it's probably fine to just delete altogether since we still have it in git history (and our hearts)

OwenPriceSkelly (approved):
cathartic diff. LGTM

scripts/migrate_search_index.py
WillEngler: Yeah, good call. I'll just delete it

PR #593, opened 2025-07-09
Clean out unused entrypoint features

WillEngler:
Draft for now


pyproject.toml
```diff
@@ -43,8 +43,6 @@ globus-compute-sdk = "<2.28.0"
 tabulate = "^0.9.0"
 types-tabulate = "^0.9.0.3"
 nbformat = "^5.9.2"
-docker = "^6.1.3"
```
OwenPriceSkelly, line 46: I think we can clean more of these -- double check me on these but I don't think we need to depend on any of the following: 
- `dill` (probably safest as a transitive globus compute dependency now) 
- `gitpython` / `huggingface-hub` (no more modelconnectors) 
- `jinja2` / `nbformat` (no more notebooks)

OwenPriceSkelly: we can also remove/bump the ceiling on the globus compute sdk itself

garden_ai/constants.py
```diff
@@ -10,8 +10,8 @@
 _DEV_ENDPOINT = "https://api-dev.thegardens.ai"
 _LOCAL_ENDPOINT = "http://localhost:5500"
 
-_PROD_ECR_REPO = "public.ecr.aws/x2v7f8j4/garden-containers-prod"
-_DEV_ECR_REPO = "public.ecr.aws/x2v7f8j4/garden-containers-dev"
+# _PROD_ECR_REPO = "public.ecr.aws/x2v7f8j4/garden-containers-prod"
```
OwenPriceSkelly, line 13: any reason not to just delete these?

OwenPriceSkelly (approved):
(forgot to hit "approve" instead of "comment" on the other one)

Issue #592, opened 2025-07-07
Batch relaxation function deployed on Edith and available in MLIP garden


WillEngler:
From the 🐮 doc:

The deployed function takes the contents of an .xyz file as argument. It uses TorchSim to efficiently run sub-batches from the user’s input and incrementally writes them to an output .xyz file. It takes the name of a particular supported MLIP checkpoint as an argument (eg, MACE-MP-0). There is a second function deployed on the GCMU endpoint that lets users retrieve the results when they’re done.

PR #610, opened 2025-07-23
Improved batch relaxation implementation

hholb:
Resolves: #592 

## Overview

This PR improves the implementation of the Edith MLIP batch relaxation flow. Major improvements:

- Use globus-compute task ids directly instead of keep state in the garden class
- Send xyz files to the remote endpoint in string chunks instead of turning them into ASE atoms locally.
- Removed ASE from the SDK dependencies
- `get_job_status` now returns a dataclass with more useful information
- Batch jobs write results to an output file incrementally on the remote endpoint
- `get_results` now checks the job status, raises an error if something went wrong during the run, and pulls down the output file and saves it locally, returning the path to the local output file.
- Bumps the globus-compute-sdk version and fixes some test fixtures that are not needed in the new version.

## Discussion

Things we still need to do/think about:

- This still doesn't have all of the model initialization code for available MLIPs in torch-sim. This includes mace variants, and the classical models - Lennard-Jones, Soft-Sphere, etc. We still need to get the setup code in-place for fairchem and other fancy models.

- Each chunk of the xyz file we send up is a new globus-compute call which means it goes through the PBS schedular on edith. This could cause a major slowdown if the queue times are long. It won't affect the relaxation as each chunk is sent in sequence and waits for the previous chunk to complete before issue a new gloubs-compute call. It is probably worth looking into a solution for sending file chunks back and forth that doesn't go through PBS and just hits the login node.

- I have a pretty basic job config for the batch jobs running on edith, see the `DEFAULT_CONFIG` in `edith_executor.py`, once this is up and running and we know it works, its probably worth taking some time to figure out a more optimal job config, or let the caller pass in a custom config.

## Testing

Manual testing of the batch processing flow.

Some unit tests for the xyz chunking implementation.

----
📚 Documentation preview 📚: https://garden-ai--610.org.readthedocs.build/en/610/


garden_ai/hpc_gardens/utils.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    xyz_path = Path(xyz_file)
+    if not xyz_path.exists():
+        raise FileNotFoundError(f"XYZ file not found: {xyz_file}")
+
+    file_size = xyz_path.stat().st_size
+    if file_size > max_size_bytes:
+        raise ValueError(
+            f"File size ({file_size / (1000*1000):.1f} MB) exceeds maximum allowed size "
+            f"({max_size_bytes / (1000*1000):.1f} MB). Please use a smaller file."
```
WillEngler, line 29: Love it

WillEngler (approved):
LGTM, thanks for iterating on this together

Issue #601, opened 2025-07-22
Batch MD functions deployed and available on Edith


hholb:
Along with relaxation, MLIPs can do MD simulations. Following the batch relaxation flow from #592, we should have a batch MD flow in the MLIP on HPC garden to capture more of what MLIPs are capable of.

PR #615, opened 2025-07-29
Create batch MD flow on Edith in mlip garden

hholb:
Resolves: #601 

## Overview

This PR adds a `batch_md` function that fits into the existing submit -> poll for status -> fetch results flow on Edith. Now users can do this:

```
import garden_ai

my_atoms = "./my_atoms.xyz"

g = garden_ai.get_garden("mlip-garden")

edith_ep_id = "<endpoint-id>"
job_id = g.batch_md(my_atoms, model='mace-mp-0', cluster_id=edith_ep_id)

status = g.get_job_status(job_id)

while status.status == "pending":
    status = g.get_job_status(job_id)

results = g.get_results(job_id)
```

Also removes some cruft from my previous PR.

## Discussion

This PR also improves the incremental output writing logic. Before we weren't actually writing outputs incrementally, we were just passing the entire batch to torch-sim and writing the processed batch to an xyz file on edith when the entire batch finished. I added an outer loop that breaks the inputs into chunks, controlled by the `max_batch_size` parameter, and then passes each chunk to torch-sim. We still use the torch-sim autobatcher for each chunk to manage the GPU memory efficiently.

I realized the model initialization, batching, and IO handling was the same for relaxation and MD, the only difference is whether we call `ts.optimize` or `ts.integrate` (and their respective args). So I factored the common logic into a `_run_batch_computation` function that takes a string arg. of either 'relaxation' or 'md' to determine which call into torch sim to make. This simplifies the structure and removes a lot of duplicate code which should make it easier to maintain going forward.

### Things we still need to do
- add better model initialization logic, support more models
- allow callers to pass options to the integrate and optimize methods for more flexible usage of the models
- allow callers to pass configuration options to the globus-compute endpoint to facilitate more control over how the job runs on the HPC

## Testing
Manual testing

----
📚 Documentation preview 📚: https://garden-ai--615.org.readthedocs.build/en/615/

### DIFF
--- a/garden_ai/hpc_gardens/mlip_garden.py
+++ b/garden_ai/hpc_gardens/mlip_garden.py
@@ -37,6 +37,40 @@ class MLIPGarden(Garden):
         )
         super().__init__(metadata=metadata)
 
+    def relax(
+        self,
+        xyz_file_path: str | Path | None = None,
+        xyz_string: str | None = None,
+        model: str = "mace",
+        options: dict = {},
+    ):
+        cloud_mlip_garden = self.client.get_garden("10.26311/qs92-3t67")
+        models_to_class_name = {
+            "mace": "MACE",
+            "orb": "ORB",
+            "fairchem": "FAIRCHEM",
+            "mattersim": "MATTERSIM",
+            "sevennet": "SEVENNET",
+        }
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
LGTM!

Issue #618, opened 2025-07-30
`garden-ai mcp setup` subcommand


OwenPriceSkelly:
It could be cool for a subcommand to automatically add the appropriate json to the user's claude desktop config (or cursor / claude code etc)

PR #623, opened 2025-08-05
Command for automating mcp config setup

loganalt9:
#618 

## Overview

This PR adds a subcommand under mcp called `setup`,  which allows for a user to automatically setup a config file. Right now this command has two potential arguments `client` and `path`. A user can pass in a client (which must be one of the five possible mcp clients this PR supports) and the config file will automatically be set up. A user can also pass in a path to a config file for a none supported mcp client and it will set up the garden mcp server in it.

Example calls:
`garden-ai mcp setup --client claude`
`garden-ai mcp setup --path path/to/config`

## Discussion

The differences in config file locations based on OS was a little interesting to get past, and definitely something to look out for. Since these paths are hard coded in any external changes about where these files live could break it for that OS. Although I think having the option to pass through a path helps keep this command stable.

## Testing

Locally on both MacOS and Windows.

----
📚 Documentation preview 📚: https://garden-ai--623.org.readthedocs.build/en/623/

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5631,6 +5631,36 @@ socks = ["pysocks (>=1.5.6,!=1.5.7,<2.0)"]
 zstd = ["zstandard (>=0.18.0)"]
 
 [[package]]
+name = "uv"
+version = "0.8.4"
+description = "An extremely fast Python package and project manager, written in Rust."
+optional = true
+python-versions = ">=3.8"
+groups = ["main"]
+markers = "extra == \"mcp\""
+files = [
+    {file = "uv-0.8.4-py3-none-linux_armv6l.whl", hash = "sha256:f9a5da616ca0d2bbe79367db9cf339cbaf1affee5d6b130a3be2779a917c14fa"},
+    {file = "uv-0.8.4-py3-none-macosx_10_12_x86_64.whl", hash = "sha256:4d8422b3058998d87fee46d4d1a437e202407cafca8b8ac69e01c6479fbe0271"},
+    {file = "uv-0.8.4-py3-none-macosx_11_0_arm64.whl", hash = "sha256:edc813645348665a3b4716a7d5e961cf7c8d1d3bfb9d907a4f18cf87c712a430"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_aarch64.manylinux2014_aarch64.musllinux_1_1_aarch64.whl", hash = "sha256:c2323e915ae562db4ebcdf5e20d3dd37a14959d07cc54939d86ab0dcdbf08f58"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_armv7l.manylinux2014_armv7l.whl", hash = "sha256:96d7a68c360383d638c283811d57558fbf7b5f769ff4bdbc99ee2a3bf9a6e574"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:385dec5a0c0909d5a24af5b02db24b49b025cbed59c6225e4c794ff40069d9aa"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_ppc64.manylinux2014_ppc64.whl", hash = "sha256:b2230310ca303328c9fd351044fb81349f3ccfaa2863f135d37bfcee707adfd1"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:86d64c66993eb0d9821caea27920175a27cd24df1eba8a340d8b3ae4074fac77"},
+    {file = "uv-0.8.4-py3-none-manylinux_2_17_s390x.manylinux2014_s390x.whl", hash = "sha256:624cf5b7bdc5cc0253115fefaad40008205d4acf34b77b294479dfe4eacb9697"},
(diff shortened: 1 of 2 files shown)


garden_ai/mcp/config_add.py
```diff
@@ -0,0 +1,100 @@
...
+    @staticmethod
+    def _initialize_config_file(config_path: Path):
+        config_path.parent.mkdir(parents=True, exist_ok=True)
+
+        if config_path.exists():
+            with open(config_path, "r") as f:
+                # If file is empty or contains invalid JSON, start with empty config
+                try:
+                    config = json.load(f)
+                except json.JSONDecodeError:
```
OwenPriceSkelly, line 84: also not sure if we want to clobber invalid json -- maybe I had a bunch of MCP servers I was copy/pasting into my config and forgot a comma. I'd be bummed if the garden server setup erased all my not-quite-valid config before I had a chance to fix it

garden_ai/mcp/config_add.py
```diff
@@ -0,0 +1,100 @@
...
+                    config = {}
+        else:
+            config = {}
+
+        if "mcpServers" not in config:
+            config["mcpServers"] = {}
+
+        config["mcpServers"]["garden-ai"] = {
+            "command": "uv",
```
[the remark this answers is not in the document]
OwenPriceSkelly: it would also mean that even uninstalling the existing `uv` wouldn't break the MCP server, since the server config would point to a `uv` executable in the directory that `garden-ai` installed to

garden_ai/app/main.py
```diff
@@ -62,6 +66,49 @@ def mcp():
...
+    elif not client and not path:
+        raise ValueError("Specify either a client or a path")
+
+    if path:
+        config_path = Init.custom(path)
+    elif client:
+        match client.lower():
+            case "claude":
+                config_path = Init.claude()
+            case "claude code":
```
OwenPriceSkelly, line 94: imo `"claude-code"` would be better than `"claude code"`, just because style wise it's a bit odd to have the CLI option expecting a space in the string

garden_ai/app/main.py
```diff
@@ -62,6 +66,49 @@ def mcp():
...
+
+    if client and path:
+        raise ValueError("Cannot specify both a client and a path")
+    elif not client and not path:
+        raise ValueError("Specify either a client or a path")
+
+    if path:
+        config_path = Init.custom(path)
+    elif client:
+        match client.lower():
```
OwenPriceSkelly, line 91: nice use of match/case

garden_ai/mcp/config_add.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+        else:  # Linux
+            config_path = (
+                Path.home() / ".config" / "Claude" / "claude_desktop_config.json"
+            )
+
+        return MCPConfigInitalizer._initialize_config_file(config_path)
+
+    @staticmethod
+    def claude_code():
```
OwenPriceSkelly, line 34: style nit: since the methods are doing something side-effectful we should give them verb names (`setup_claude_code`) instead of just noun names. Without having read the code I'd probably have guessed that the `claude_code()` method just finds and returns the path to the config, not modify it

garden_ai/mcp/config_add.py
```diff
@@ -0,0 +1,100 @@
...
+        command = "claude mcp add garden -- uv run --with garden-ai[mcp] garden-ai mcp"
+
+        result = subprocess.run(command.split(" "), capture_output=True, text=True)
+
+        if result.stdout:
+            # Parse claude response for file path
+            parts = result.stdout.split("File modified: ")
+            if len(parts) > 1:
```
OwenPriceSkelly, line 42: I feel like this also wants to be a match/case 
```
match result.stdout.split():
    case "File", "modified:", *parts:
        return pathlib.Path(*parts) 
    case _:
        raise HelpfulErrorType(...)
```

garden_ai/mcp/config_add.py
```diff
@@ -0,0 +1,100 @@
...
+
+        return MCPConfigInitalizer._initialize_config_file(config_path)
+
+    @staticmethod
+    def custom(path: str):
+        return MCPConfigInitalizer._initialize_config_file(Path(path))
+
+    @staticmethod
+    def _initialize_config_file(config_path: Path):
+        config_path.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
OwenPriceSkelly: we should also ensure the path is absolute (I believe `.resolve()` is the method)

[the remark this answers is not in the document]
loganalt9: In retrospect I completely agree with this. I did this because one of the hard coded mcp clients (I think it was gemini) didn't automatically set up the config file on installation, but silently creating the file is not the best way to go about it. Probably better to raise an error that shows where the config file should exist

garden_ai/mcp/config_add.py
loganalt9: I think there might be an issue with this due to claude code outputting another line of text before this. From my testing any match/case like this would always evaluate to the default case, since the 'case' in this context for `result.stdout.split()` as a whole would be `"Added", "stdio", *parts`

What claude code outputs after successfully adding an mcp server:
```
Added stdio MCP server garden with command: uv run --with garden-ai[mcp] garden-ai mcp to local config
File modified: /Users/loganalt/.claude.json [project: /Users/loganalt/dev/garden]
```

OwenPriceSkelly (approved):
LGTM! good stuff 🙌 🌱

PR #632, opened 2025-08-12
Edith fix: trick globus compute into creating new endpoints for each job

hholb:
## Overview

This PR fixes an issue we uncovered last week where subsequent batch relaxation jobs failed with confusing errors about missing files on the remote compute node.

This also bumps the batch sizes down a little to avoid out of memory errors SevenNet was hitting on large structures.

A quick summary of what this PR is meant to do.

## Discussion

So this was a bit tricky to solve. The bug we were seeing was triggered when calling `batch_relax` more than once in quick succession in the same python session. The first call would run fine, but subsequent calls would fail with errors like `Error: FileNotFound /var/PBS/<job-id>/generated_script.py` where the PBS job id would be the job id from the first job. Turns out this was as subtle behavior of globus-compute seeing that the PBS job from the first call was still alive and reusing it for subsequent calls. I think this is an optimization that makes sense when running short-lived functions to avoid scheduler queue times. But for our patterns we want subsequent calls to start new PBS jobs. The "fix" is to trick globus-compute to think each request has a "different" `worker_init` script which causes the multi-user endpoint to start a new user endpoint for each request, which in turn starts a new PBS job for each invocation. I do this by adding a comment string to the worker init script with a timestamp that doesn't do anything, but does make globus-compute think its a different endpoint config.

## Testing

Manual testing of multiple `batch_relax` calls on different models in the same python session while watching the PBS queue on Edith. I saw three separate jobs that all succeeded!

### DIFF
--- a/garden_ai/hpc_executors/edith_executor.py
+++ b/garden_ai/hpc_executors/edith_executor.py
@@ -11,7 +11,7 @@ from garden_ai.hpc_gardens.utils import subproc_wrapper  # noqa:
 EDITH_EP_ID = "a01b9350-e57d-4c8e-ad95-b4cb3c4cd1bb"
 
 DEFAULT_CONFIG = {
-    "worker_init": """true;
+    "worker_init": """
 module load openmpi
 export PATH=$PATH:/usr/sbin
 """,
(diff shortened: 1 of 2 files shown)


garden_ai/hpc_executors/edith_executor.py
```diff
@@ -26,6 +26,16 @@ def __init__(
         user_endpoint_config=DEFAULT_CONFIG,
         **kwargs,
     ):
+        # vary the user_endpoint_config to try to force separate PBS jobs
+        if "user_endpoint_config" in kwargs and kwargs["user_endpoint_config"]:
+            import time
+
+            config = kwargs["user_endpoint_config"].copy()
+            # Add a unique comment to make each job "different"
```
WillEngler, line 34: baller

WillEngler (approved):
LGTM!

Issue #629, opened 2025-08-11
migrate from `poetry`->`uv`


OwenPriceSkelly:
We've been drinking the `uv` kool-aid for a while now, but we're still using `poetry` for sdk development. Let's get with the times: 

  - [x] `pyproject.toml` should use the [PEP-621](https://peps.python.org/pep-0621/) standardized syntax for dependencies instead of the poetry-specific version specifiers 
  - [x] our `pre-commit` hooks should use the official uv hook to keep pyproject and `uv.lock` in sync 
    - while we're at it, we might as well move from `black`/`isort` etc to `ruff`
    - The backend repo's [pre-commit config](https://github.com/Garden-AI/garden-backend/blob/dev/.pre-commit-config.yaml) is a good reference  
  - [x] our regular CI action should install dependencies and run tests using uv 
  - [x] our publish to pypi action should use uv to build and publish the package 
  - [x] our readthedocs config will need to use uv instead of poetry
  - [ ] misc docs / readmes that mention poetry should be about uv instead

PR #633, opened 2025-08-12
Migrate from poetry to uv

loganalt9:
#629 

## Overview

This PR migrates from Poetry to uv by updating all of the relevant config files such as `pyproject.toml`, `.readthedocs.yaml`, `.pre-commit-config.yaml`, `ci.yaml`, `pypi.yaml`. This PR also switches over from using black/isort to ruff as our linter. Some processes including our github CI and publish to pypi have been updated to install dependencies and run with uv. Developer docs also updated to mention the correct uv calls instead of poetry.

## Discussion

Two small areas in this PR that don't feel too strong are in `.readthedocs.yaml` and `garden_ai/app/main.py`. 

For `readthedocs.yaml` I couldn't find a solution to get the RTD environment synced up with uv while building, so here I just naively pip installed the necessary packages, more than open to suggestions on a better solution for this.

For `garden_ai/app/main.py` I started to run into a weird bug where the test `test_print_usage_info_if_no_args` would fail. From what I gathered, with how we previously instantiated our app (`app = typer.Typer(no_args_is_help=True`) there were common issues with identifying subcommands when you start adding subcommands with`add_typer` (new with the mcp setup command). To get around this I manually check for subcommands in a call, but also more than open to better solutions for this.

TL;DR: #629 checklist covers all that this PR does

## Testing

Passing all tests, also running normally with uv sync/uv run flow now.  

## Documentation

Updated documentation from poetry to uv

----
📚 Documentation preview 📚: https://garden-ai--633.org.readthedocs.build/en/633/


.readthedocs.yaml
```diff
@@ -12,15 +12,11 @@ build:
...
-      - pip install poetry==1.8.2
+      - pip install uv
     post_install:
-      - VIRTUAL_ENV=$READTHEDOCS_VIRTUALENV_PATH poetry install --with develop
-      # Debugging CI build
-      - pip freeze
-      - mkdocs --version
-      - python -m site
-      - poetry run mkdocs --version
+      - uv sync --extra develop
```
OwenPriceSkelly, line 17: I think that if we have the same `VIRTUAL_ENV=$READTHEDOCS_VIRTUALENV_PATH` variable set, `uv sync` should install the requirements to the right location without needing the additional `pip install`s

OwenPriceSkelly (approved):
LGTM! thanks for knocking this out 🙌

Issue #637, opened 2025-09-23
Fetch HPC function data from backend to call via globus-compute


hholb:
Once HPC functions are stored in the backend database and we have routes for fetching the data we need to figure out how to munge the data so that we can call the functions through our gloubs-compute setup.

Issue #638, opened 2025-09-23
Send HPC invocation logs to the backend


hholb:
Once HPC invocations are working in the SDK, we need to send some tracking info to the backend so we can log metrics on HPC invocations. The route for doing so has not been created yet, but will likely be something like `POST /hpc-functions/{function_id}`

PR #639, opened 2025-10-02
Add support for HPC functions attched to gardens from the backend

hholb:
Resolves: #637 

## Overview

This PR add schemas to build client-side representations of HPC functions and refactors the `HpcExectuor` (formerly `EdithExectuor`) to call any hpc function attached to garden via globus-compute. It doesn't send any invocation logs to the backend yet -- that will come in #638.

The general flow from a user's view looks like this:
```python
import garden_ai
g = garden_ai.get_garden("some-doi")

# see if there are any hpc functions
g.hpc_functions

# for an hpc funcion, see what endpoints it can run on
g.hpc_func.endpoints # for example, prints: {name: "ALCF Edith", id: "globus-uuid"}

# run an hpc function on an endpoint, return value is a globus-compute task id
job_id = g.hpc_func.submit(*args, endpoint_id="some-uuid", **kwargs)

# poll for status
g.get_job_status(job_id)

# fetch results
g.get_job_results(job_id)
```

## Discussion
I updated the `_subproc_wrapper` helper to remove the hard-coded conda env path for the mlip garden. It now takes a `conda_env_path` argument that comes down from the backend with the deployment info associated with an hpc function.

## Testing
Attempted manual testing of MLIP garden, but Edith MEP is down.

New unit tests

## Documentation

No docs updates here. But we should definitely update the docs with examples once the API is stable.


garden_ai/hpc/functions.py
```diff
@@ -0,0 +1,113 @@
...
+                    "name": dep_info.endpoint_name,
+                    "id": endpoint_id,
+                }
+        self.endpoints = list(seen_endpoints.values())
+
+    def __call__(self, *args, **kwargs):
+        raise NotImplementedError(
+            f"HPC functions must be submitted asynchronously.\n"
+            f"Try:\n\tjob_id = {self.metadata.function_name}.submit("
+            f"endpoint_id='{self.endpoints[0]['id'] if self.endpoints else 'ENDPOINT_ID'}', "
```
OwenPriceSkelly, line 34: nit: think this suggestion from the error message might be not quite valid python since positional `*args` would have to come before the `endpoint_id` kwarg

garden_ai/backend_client.py
```diff
@@ -59,11 +59,11 @@ def _delete(self, resource: str, payload: dict) -> dict:
     def _get(self, resource: str, params: dict | None = None) -> dict:
         return self._call(requests.get, resource, None, params=params)
 
-    def get_garden(self, doi: str) -> Garden:
+    def get_garden(self, doi: str, client=None) -> Garden:
```
OwenPriceSkelly, line 62: this smells a little odd, iirc the only reason `_from_nested_metadata` even accepts a client was a kludge for unit tests. since the GET request to the backend only needs a doi, I think the BackendClient.get_garden should also only take a doi

OwenPriceSkelly (changes requested):
looking good outside of a couple of nits -- I think we already have an awkward relationship between `Client` and `BackendClient`, and I think it's worth keeping them as separate as possible, unless we're willing to fully refactoring them into one class (which I think is probably not worth the squeeze)

garden_ai/hpc/functions.py
hholb: Good catch!

PR #641, opened 2025-10-03
Send hpc invocation logs to the backend

hholb:
Resolves: #638 

## Overview

This PR adds a request schema and logic to send invocation logs to the backend's `POST /hpc/invocations` route when user call `submit` on an HPC function


OwenPriceSkelly (approved):
LGTM

Issue #646, opened 2025-11-07
Bump groundhog version


hholb:
There have been quite a few nice additions to groundhog, we should update to the latest version.

Issue #645, opened 2025-11-07
Update HPC schemas


hholb:
The backend added a new field to the HPC endpoint schema, we need to add it here to be compatible.

PR #647, opened 2025-11-07
Fix HPC endpoint shcema, bump globus-compute version

hholb:
Resolves: #645 
Resolves: #646 

## Overview

This PR fixes the HPC endpoint schema to be a dict rather than just a string now that the backend sends down the name and uuid of the endpoint. This was causing a pydantic error trying to build the `HpcFunction` class from metadata we fetch from the backend.

Also bumps the globus-compute-sdk and groundhog packages to the most recent versions. Along with the groundhog update is the addition of the `.local()` method for HPC functions.

Removed jupyter from dev dependencies to resolve a new CVE.

## Testing

Manually

_Before_
image

_After_
image

----
📚 Documentation preview 📚: https://garden-ai--647.org.readthedocs.build/en/647/

### DIFF
--- a/garden_ai/hpc/groundhog.py
+++ b/garden_ai/hpc/groundhog.py
@@ -23,7 +23,7 @@ def load_function_from_source(contents: str, name) -> Function:
         script_path = Path(temp_file.name)
         with groundhog_script_path(script_path):
             module = import_user_script(f"{name}_module", script_path)
-            obj = getattr(module, name)
+            obj = module.__dict__.get(name)
             # import __main__
             #
             # # exec the script to init the Function instance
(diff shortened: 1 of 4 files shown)


garden_ai/hpc/groundhog.py
```diff
@@ -21,18 +22,20 @@ def load_function_from_source(contents: str, name) -> Function:
...
-            if obj is None:
-                raise ValueError(f"No groundhog function {name} found in script")
-            elif not isinstance(obj, Function):
-                raise ValueError(
-                    f"Expected {name} to be a groundhog function, got: {type(obj)}"
-                )
+            module = import_user_script(f"{name}_module", script_path)
+            obj = module.__dict__.get(name)
+            # import __main__
+            #
```
WillEngler: This seems like a functionality change and not a schema change? (I am not deep enough in recent developments to tell if this makes sense or not, sadly)

OwenPriceSkelly (approved):
LGTM - gonna merge now to cut a release 🚢

